PR #1, opened 2023-02-15
Initial Design commit with todos

coax1d:
Current check list of Todos (Totally open and willing to change this):
- [x] Initial design structure for different kinds of transactions
- [x] Working flow of different UTXOS
- [x] Implement `TuxedoPiece` trait on each user defined UTXO type
    - [x] implement `fn validate` for MoneyPiece
        - [x] Tests for MoneyPiece
    - [x] implement 'fn validate' for KittiesPiece
        - [ ] Tests for KittiesPiece
    - [ ] implement 'fn validate' for ExistencePiece
        - [ ] Tests for ExistencePiece
- [ ] Implement `UtxoSet` trait on each UTXO type i.e. KittiesPiece, MoneyPiece, etc..  or just on a single `UtxoSet` type?
- [ ] Runtime Tests for overall system
- [ ] Refactor and cleanup all old lingering code which is unecessary
- [ ] Tuxedo version of `construct_runtime!`? So that we can for instance configure the `Call` enum for different TuxedoPieces
- [ ] Integration Test in a multi UTXO Tuxedo Runtime
- [ ] ChainSpec and CLI support for Tuxedo

This most likely will not be a single pr but is more of a brain dump of tasks.

### DIFF
--- a/frameless-runtime/src/utxo.rs
+++ b/frameless-runtime/src/utxo.rs
@@ -101,7 +101,7 @@ pub trait UtxoSet {
     /// Insert the given utxo into the state storing it with the given ref
     /// The ref is probably the hash of a tx that created it and its index in that tx, but this decision is opaque to this trait
     /// Return whether the operation is successful (It can fail if the ref is already present)
-    fn insert(utxo_ref: UtxoRef, utxo: Utxo) -> bool;
+    fn insert(utxo_ref: UtxoRef, utxo: &Utxo) -> bool;
 
     ///
     /// nullify the utxo by either:
@@ -129,9 +129,10 @@ pub trait TuxedoPiece {
     /// The type of data stored in Outputs associated with this Piece
     type Data: Encode + Decode;
     const TYPE_ID: TypeId;
+    type Error: Default;
 
     /// The validation function to determine whether a given input can be consumed.
-    fn validate(transaction: Transaction) -> bool;
+    fn validate(transaction: Transaction) -> Result<(), Self::Error>;
 }
 
 pub struct PieceExtracter<Piece>(PhantomData<Piece>);
@@ -139,6 +140,10 @@ impl<Piece: TuxedoPiece> PieceExtracter<Piece> {
     pub fn extract(key: UtxoRef) -> Result<Piece::Data, ()> {
         let encoded_utxo = sp_io::storage::get(&key.encode()).ok_or(())?;
         let utxo = Utxo::decode(&mut &encoded_utxo[..]).map_err(|_| ())?;
+        Self::extract_from_output(&utxo)
+    }
+
+    pub fn extract_from_output(utxo: &Utxo) -> Result<Piece::Data, ()> {
         if utxo.data_id != Piece::TYPE_ID {
             return Err(())
         }
@@ -155,7 +160,7 @@ impl UtxoSet for MoneyPiece {
         sp_io::storage::exists(&utxo_ref.encode())
     }
 
-    fn insert(utxo_ref: UtxoRef, utxo: Utxo) -> bool {
+    fn insert(utxo_ref: UtxoRef, utxo: &Utxo) -> bool {
         sp_io::storage::set(&utxo_ref.encode(), &utxo.encode());
         true
     }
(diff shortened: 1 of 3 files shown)


frameless-runtime/src/utxo.rs
```diff
@@ -86,25 +110,91 @@ pub trait UtxoSet {
...
-        // decode the transaction output data as type `Data`
+        // Check that the input is unique and a set
+        {
+            let input_set: BTreeSet<_> = transaction.inputs.iter().collect();
+            if input_set.len() < transaction.inputs.len() {
+                return false;
+            }
+        }
+
+        // 1.) Check that the outputs referenced by each input can be redeemed
```
JoshOrndorff, line 187: This is the biggest disadvantage I see with the way your Call enum is structured. Having the utxo-style transaction embedded in each call variant means that each tuxedo piece will have to include this logic. We, the tuxedo core developers, could make this a little nicer for piece developers by providing a standard helper function that they can just call At least then they don't have to re-write it or copy paste it, but there is still the possibility that they forget to call it entirely.

coax1d: You are right can probably wrap it up nicely in a function or somehow make it default always done no matter what. I just see a path ahead and was going with it. I can swap to another paradigm if we want.

frameless-runtime/src/utxo.rs
```diff
@@ -0,0 +1,253 @@
...
+
+/// TODO: Clean up this file and organize different parts into different modules for easier reading.
+
+///
+/// TODO: Something similar to construct_runtime! which will setup all the configurations for a UTXO runtime
+/// construct_utxo_runtime!(
+///     MoneyUTXO // Configuration for MoneyUTXO
+///     KittiesUTXO // Configuration for KittiesUTXO
+///     ExistenceUTXO // Configuration for ExistenceUTXO
+/// )
```
JoshOrndorff, line 28: I definitely agree with this. I think the idea of aggregating pallets/pieces into runtime-level "outer" data types is worth copying from FRAME.

frameless-runtime/src/utxo.rs
```diff
@@ -0,0 +1,253 @@
...
+    fn redeem(self, tx: &[u8], witness: &[u8]) -> bool {
+        let signature = match Signature::try_from(&witness[..]) {
+            Ok(sig) => sig,
+            Err(_) => return false,
+        };
+        sp_io::crypto::sr25519_verify(&signature, &tx, &Public::from_h256(self))
+    }
+}
+
+// TODO: Implement this for Each Tuxedo Piece
```
JoshOrndorff, line 93: We should think through the advantages to having a separate utxo set for each piece as opposed to a single utxo set. I'm not currently seeing advantages to having more than one, but I could just be missing them.

frameless-runtime/src/utxo.rs
```diff
@@ -0,0 +1,253 @@
...
+        match Utxo::decode(&mut &encoded_utxo[..]) {
+            Ok(utxo) => Some(utxo),
+            Err(_) => None,
+        }
+    }
+}
+
+pub trait Get<T> {
+    fn get(&self) -> T;
+}
```
JoshOrndorff, line 125: There are a few things, like this for example, that are currently in `frame_support`, but are useful outside of frame. Another one I encountered is `ensure!`. Once we have a sense of what these things are, we should PR Substrate to move them to a less frame-specific place. Maybe it warrants a new crate called something like `runtime_support`.

frameless-runtime/src/utxo.rs
```diff
@@ -0,0 +1,253 @@
...
+    /// For the Money UTXO we just want to consume it.
+    fn nullify(utxo_ref: UtxoRef) -> Option<Utxo> {
+        let encoded_utxo = sp_io::storage::get(&utxo_ref.encode())?;
+        sp_io::storage::clear(&utxo_ref.encode());
+        match Utxo::decode(&mut &encoded_utxo[..]) {
+            Ok(utxo) => Some(utxo),
+            Err(_) => None,
+        }
+    }
+}
```
JoshOrndorff, line 172: none of this is money-specific. Makes me think we should not have to impl this trait for each piece.

frameless-runtime/src/utxo.rs
```diff
@@ -0,0 +1,253 @@
...
+
+impl TuxedoPiece for MoneyPiece {
+    type Data = u128;
+    const TYPE_ID: TypeId = *b"1111";
+
+    fn validate(transaction: Transaction) -> bool {
```
JoshOrndorff, line 178: This is the thing I'm liking most about your design. This is like a combination of my `Verifier` and `UtxoData` traits. I'm starting to think combining them makes a lot of sense.

I started to refactor toward this, but didn't make it too far (because it's late, not because of fundamental problems). I also added a `type Error` in mine.

frameless-runtime/src/utxo.rs
```diff
@@ -0,0 +1,253 @@
...
+    /// Return whether the operation is successful (It can fail if the ref is already present)
+    fn insert(utxo_ref: UtxoRef, utxo: Utxo) -> bool;
+
+    ///
+    /// nullify the utxo by either:
+    /// - Consuming it entirely
+    /// - Putting it on "Timeout"
+    /// - Not consuming it but marking it as spent
+    ///
+    fn nullify(utxo_ref: UtxoRef) -> Option<Utxo>;
```
JoshOrndorff, line 112: I like the flexibility you're planning for here. We might need to add another parameter to the `nullify` function (possibly with a generic type). For example, in the ZK utxo set, we would have to provide a nullifier (like publishing the serial number) right? I guess in the simple case of a totally transparent utxo set that type could just be `()`. Just thinking out loud here.

frameless-runtime/src/utxo.rs
coax1d: I initially felt this way but I thought perhaps the nullify part might have different implementations depending on the Piece. I couldnt think of any at the time but just because I couldnt think of any I didnt want to assume it wasnt possible.

frameless-runtime/src/utxo.rs
coax1d: I did notice there are several little things in frame that are really not tied down to frame at all.

frameless-runtime/src/utxo.rs
coax1d: Yes to the second comment and first comment I agree can rework this.

frameless-runtime/src/utxo.rs
coax1d: This is what I would like to converge on. I would gladly take on your design I think it is really solid only thing I want to change is merging into One place where you have the Data Type of a particular UTXO and how to validate or verify a transition for this particular utxo type

frameless-runtime/src/utxo.rs
coax1d: Would like to explore this thought a bit more again together then ask Jeff about it

JoshOrndorff:
This has been merged to main through the cli.

PR #2, opened 2023-02-15
Braindump for initial design

JoshOrndorff:
This starts to capture my swirling ideas regarding the data structures, the redeemer / verifier distingtion, and how piece aggregation might work. Admittedly, there is still a lot to be solidified, but it is a place to start chatting.


JoshOrndorff:
Superceded by #3

PR #3, opened 2023-02-24
Proposed Main

JoshOrndorff:
This work combines the ideas from #1 and #2 to establish a tuxedo core as well as an example node that includes several pieces.

This should NOT be squashed because we want to keep all the history. In fact, if we agree to it, we should just push it to `main` with the cli rather than using the github ui.


JoshOrndorff:
Most of this was merged through the CLI. I'll close this and the remaining work here can be a separate PR.

PR #5, opened 2023-02-27
Amoeba tests

JoshOrndorff:
This PR primarily adds tests for the amoeba example piece.
It also fixes one small error that the tests found.

And finally, it moves the `Bogus` type from the money piece to a reusable location in the tuxedo core crate so that it can be used in tests for all pieces.

### DIFF
--- a/frameless-runtime/src/amoeba.rs
+++ b/frameless-runtime/src/amoeba.rs
@@ -37,7 +37,7 @@ impl UtxoData for AmoebaDetails {
 }
 
 /// Reasons that the amoeba verifiers may fail
-#[derive(Debug)]
+#[derive(Debug, Eq, PartialEq)]
 pub enum VerifierError {
     // TODO In the current design, this will need to be repeated in every piece. This is not nice.
     // Well, actually, no. Some pieces will be flexible about how many inputs and outputs they can take.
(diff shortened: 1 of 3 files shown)


frameless-runtime/src/amoeba.rs
```diff
@@ -183,3 +183,307 @@ impl Verifier for AmoebaCreation {
...
+        let to_spawn = AmoebaDetails {
+            generation: 0,
+            four_bytes: *b"test",
+        };
+        let input_data = Vec::new();
+        let output_data = vec![to_spawn.clone().into(), to_spawn.into()];
+
+        assert_eq!(
+            AmoebaCreation.verify(&input_data, &output_data),
+            Err(VerifierError::WrongNumberOutputs),
```
coax1d, line 259: Is it possible to make this be a little more granular or explain why it is a wrong number of outputs. This may be a nit and if you understand the amoeba in detail you should know why it is the wrong number of outputs. I am thinking of something like "TooManyChildren" or something like this. If it doesnt work in this context than can just disregard.

frameless-runtime/src/amoeba.rs
```diff
@@ -183,3 +183,307 @@ impl Verifier for AmoebaCreation {
...
+        let mother = AmoebaDetails {
+            generation: 1,
+            four_bytes: *b"test",
+        };
+        let d1 = AmoebaDetails {
+            generation: 2,
+            four_bytes: *b"test",
+        };
+        let input_data = vec![mother.into()];
+        // There is only one daughter when there should be two
```
coax1d, line 379: Similar to this comment I like this comment because it explains why you have this error type which can be seen in multiple types of scenarios.

coax1d (approved):
I think for now is good the nit is I necessary for this PoC. Looks Good!

JoshOrndorff:
I think it was a good point. I've made the error more granular now. I'll still take the approval though ;)

PR #6, opened 2023-02-27
Cleanup and explain consensus authority sets

JoshOrndorff:
Tuxedo does not yet have any real PoS- or even PoA- style authority set management pieces. So for now we use simple hard-coded authority sets.

This PR improves this situation slightly by making all of the standard (Alice, Bob, Charlie, etc) keys readily available and providing comments explaining that that's what is happening.

### DIFF
--- a/frameless-runtime/src/lib.rs
+++ b/frameless-runtime/src/lib.rs
@@ -5,6 +5,7 @@ include!(concat!(env!("OUT_DIR"), "/wasm_binary.rs"));
 
 use parity_scale_codec::{Decode, Encode};
 use sp_consensus_aura::sr25519::AuthorityId as AuraId;
+use sp_finality_grandpa::AuthorityId as GrandpaId;
 
 use log::info;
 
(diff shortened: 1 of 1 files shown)


frameless-runtime/src/lib.rs
```diff
@@ -298,6 +299,61 @@ impl Verifier for OuterVerifier {
...
+
+    /// Aura authority IDs
+    fn aura_authorities() -> Vec<AuraId> {
+        use hex_literal::hex;
+        use sp_application_crypto::ByteArray;
+
+        [
+            // Alice
+            hex!("d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d"),
+            // Bob
```
coax1d, line 316: Should we leave these commented out or add them in?

JoshOrndorff: I had left them commented with `--dev` nodes in mind. If we have all six authorities enabled, then a `--dev` node (which has Alice's keys) will only author one block in every six slots.

coax1d: ok great!

coax1d:
Looks good!

PR #7, opened 2023-02-28
Adding externalities and genesis

coax1d:
Just adding the externalities in runtime for testing and the genesis money utxo test.


JoshOrndorff (approved):
Looks good to me mod one comment about the name Alice.

PR #8, opened 2023-02-28
Initial kitties refactor commit

coax1d:
Adding kitties example piece


coax1d:
The tests are still todo I think might want to just merge this and then implement tests after I work on the wallet. What do you think @JoshOrndorff

coax1d:
Going to create seperate PR for the tests and validation but prefer to break it up it is quite big

tuxedo-template-runtime/src/kitties.rs
```diff
@@ -1,126 +1,544 @@
...
+    NewChildHasNonZeroBreedings,
+    /// New child parent info is either in Tired state or HadBirthRecently state which is not possible.
+    NewChildIncorrectParentInfo,
+    /// Too many breedings for this kitty can no longer breed.
+    TooManyBreedingsForKitty,
+    /// Not enough free breedings available for these parents.
+    NotEnoughFreeBreedings,
+}
+
+trait Breed {
```
JoshOrndorff, line 131: Why is this a trait instead of just methods on the `FreeKittyVerifier` or even just helper functions. Do you imagine there will be another implementation at some point?

tuxedo-template-runtime/src/kitties.rs
```diff
@@ -1,126 +1,544 @@
...
+    ///     - Both parents free breedings is non-zero
     ///
-    /// nullify the utxo by either:
-    /// - Consuming it entirely
-    /// - Putting it on "Timeout"
-    /// - Not consuming it but marking it as spent
+    fn check_free_breedings(mom: &KittyData, dad: &KittyData) -> Result<(), Self::Error> {
+        let mom_breedings = mom.free_breedings;
+        let dad_breedings = dad.free_breedings;
+        if (mom_breedings == 0) && (dad_breedings == 0) {
```
JoshOrndorff, line 224: Should this be `||`? Doesn't it count as a breeding for each of them?

tuxedo-template-runtime/src/kitties.rs
```diff
@@ -1,126 +1,544 @@
-//! An NFT game inspired by cryptokitties.
```
JoshOrndorff, line 1: A few sentences about the general flow of the kitty breeding would be helpful. For example, are kitties inherently gendered or can the same kitty act as mom sometimes and dad other times. Maybe explain the tired period, etc.

JoshOrndorff (approved):
Overall looking good. Is this one waiting on #41 or can it go in right away?

coax1d:
> Overall looking good. Is this one waiting on #41 or can it go in right away?

This can go right away

tuxedo-template-runtime/src/kitties.rs
coax1d: True great catch hopefully tests would have gotten it

tuxedo-template-runtime/src/kitties.rs
coax1d: It is possible I was leaving the possibility open. Also was nice for having some associate types encapsulated

PR #10, opened 2023-03-01
Add Dockerfile

JoshOrndorff:
This PR adds a dockerfile for the tuxedo template node

After hours of trying to write this from first principles, it turned out that the example in the Substrate Node Template was exactly what we needed, so this is nearly a direct copy paste.

This is not yet built or published in CI. We can decide whether we want that after things are a little more stable.

To test this
```bash
git checkout joshy-dockerfile
docker build --tag tuxedo .
docker run tuxedo --version
docker run tuxedo --dev
```


coax1d (approved):
👌👍

PR #12, opened 2023-03-02
CI: Caching

JoshOrndorff:
This PR strives to add caching to the CI. I've just blindly installed the action which supposedly has "sensible defaults". Let's see whether it reduced CI time.


JoshOrndorff:
Seems like we didn't get a cache hit on this first commit. I'll merge master in a few times over the next days and see if it makes a difference.

JoshOrndorff:
Woah! Finished in < 5 min! :fire: 

I'm starting to think this is worth merging, at least to see how often we get useful cache hits.

PR #34, opened 2023-03-03
Hacking on wallet idea

JoshOrndorff:
This is a continuation of the wallet work that was already merged in #4 and #32.

Starts working toward a clap interface.


wallet/src/main.rs
```diff
@@ -1,119 +1,126 @@
...
+    /// They must all be coins and, for now, must all be owned by the same signer.
+    #[arg(long, short)]
+    input: Vec<String>,
+
+    /// Hex encoded address (sr25519 pubkey) of the recipient (non 0x prefixed for now)
+    /// Generate with subkey, or use Shawn's: d2bf4b844dfefd6772a8843e669f943408966a977e3ae2af1dd78e0f55f4df67
+    #[arg(long, short, default_value_t = SHAWN_PUB_KEY.to_string())]
+    recipient: String,
+
+    // The `action = Append` allows us to accept the same value multiple times.
```
coax1d, line 79: Great documentation love the detail

wallet/src/main.rs
```diff
@@ -1,119 +1,126 @@
...
+    }
 }
 
-async fn get_coin_from_storage(
+/// Fetch an output from chain storage given an OutputRef
+pub async fn fetch_storage<R: Redeemer>(
     output_ref: &OutputRef,
     client: &HttpClient,
-) -> anyhow::Result<(H256, Coin)> {
+) -> anyhow::Result<Output<R>> {
```
coax1d, line 112: This is more of what I was going for initially but wasnt running short on time I like this instead

wallet/src/money.rs
```diff
@@ -0,0 +1,112 @@
...
+};
+
+/// Create and send a transaction that spends coins on the network
+pub async fn spend_coins(client: &HttpClient, args: SpendArgs) -> anyhow::Result<()> {
+    println!("The args are:: {:?}", args);
+    let (provided_pair, _) = Pair::from_phrase(&args.seed, None)?;
+
+    // Construct a template Transaction to push coins into later
+    let mut transaction = Transaction {
+        inputs: Vec::new(),
```
coax1d, line 27: Only nit I would say is sometimes in the code we initialize with ‘Vec::new()’ and other times with ‘vec![]’ we should probably choose one for consistency

wallet/src/money.rs
```diff
@@ -0,0 +1,112 @@
...
+    let params = rpc_params![genesis_spend_hex];
+    let genesis_spend_response: Result<String, _> =
+        client.request("author_submitExtrinsic", params).await;
+    println!(
+        "Node's response to spend transaction: {:?}",
+        genesis_spend_response
+    );
+
+    // Wait a few seconds to make sure a block has been authored.
+    sleep(Duration::from_secs(3));
```
coax1d, line 75: Is sleep the best option we have in the long run? If anything then maybe we just sleep for the block time? Or wait for a response that the transaction is has been included?

coax1d (approved):
Looks great awesome refactoring and love the documentation I just threw in a couple minor comments. Great start to the official wallet

wallet/src/money.rs
JoshOrndorff: There must be better options. I think there are subscription based rpc endpoints. Haven't looked into that yet. sleep is for sure not great here. 3 seconds is the block time. We could import the value from the runtime crate, but I hope we'll ditch sleep in the next few weeks, so I just hard coded it.

wallet/src/money.rs
JoshOrndorff: Agreed. Do you have a preference?

Issue #37, opened 2023-03-04
Duplicate output check is unnecessary and incorrect


JoshOrndorff:
Currently the tuxedo executive checks that there are no duplicate outputs in a given transaction. https://github.com/Off-Narrative-Labs/Tuxedo/blob/main/tuxedo-core/src/executive.rs#L53-L60

This logic is also present in the earlier utxo workshop. https://github.com/JoshOrndorff/utxo-workshop/blob/joshy-update-deps-may-2022/runtime/src/utxo.rs#L188-L191

But why is it necessary? It is perfectly valid for a transaction to consume a 20 token input and create two new outputs each worth 5 tokens that belong to the same pubkey.

Digging back through the git history, it seems this check has been present since a much older 2019 edition of the utxo-workshop where each output contained a [user-provided salt](https://github.com/substrate-developer-hub/utxo-workshop/blob/63ffc63191edde1c124fdcf715a65e67b089402a/runtime/src/utxo.rs#L61-L64). The user-provided salt was never realistic and was a hack to get something working in the early days. @nczhu noticed this and solved it in https://github.com/substrate-developer-hub/utxo-workshop/pull/45 (still an impressive PR after all this time!).

With the salt hack long-gone, I think it is safe to remove the duplicate output check.

coax1d:
I remember doing this in my initial version during academy one. When I presented at Academy 2 I noted that you can have two outputs which are exactly the same and when they are stored in the tree they are still guaranteed to have different keys.

coax1d:
Fixed with #38

PR #38, opened 2023-03-04
Remove duplicate output check

JoshOrndorff:
Solves #37 by removing the unnecessary duplicate output check.


coax1d (approved):
Looks good as discussed in #37 👍

Issue #27, opened 2023-03-02
Reconsider the strong `Redeemer` / `Verifier` separation


JoshOrndorff:
In the current design, Tuxedo has a very strong separation between the logic that allows a single input to be consumed and the logic that checks whether a transaction is valid. This issue is to consider the pros and cons of that separation and decide to what extent it should be kept.

## Note on Terminology

Before going on, I'll observe that I _accidentally_ used the same terms as [IOHK's abstract model](https://eprint.iacr.org/2018/469.pdf) with different meanings :facepalm:

| Tuxedo Term | Abstract Model Term | Definition |
| ----------- | ------------------- | ---------- |
| Verifier    | None - token logic hard coded | The logic used to determine whether the consumed inputs and created outputs meet all necessary constraints |
| Redeemer    | Verifier            | The logic used to determine whether an individual input can be consumed |
| Witness     | Redeemer            | The data, likely a signature, that satisfies the logic to consume an input

### Pro 1: Writing pieces is very simple

With the current separation in place, the author of a piece doesn't need to care anything about whether individual inputs can be consumed. This is entirely handled by tuxedo executive. By the time the execution reaches the individual piece all that is left is typed data for the piece to check.

### Con 1: Pieces can't forcefully consume inputs

In some cases it is useful to consume a UTXO for housekeeping purposes.

Example 1: Consider a coupon with an expiration date. This UTXO would likely be guarded by a `CheckSig` so only the owner can use it. But if the expiration date passes, anyone should be able to clean up the state.

Example 2: In the proof of existence piece, two users could claim the same hash. When these competing claims are discovered anyone should be able to clean up the latter claim because the former is more legitimate. This is encountered in the [current codebase](https://github.com/Off-Narrative-Labs/Tuxedo/blob/7f1dabe37ce8621deb9df47d102ea775bbe2f763/frameless-runtime/src/poe.rs#L149-L180)

This can be addressed with a notion of evictions. See below.

### Con 2: Dynamic ownership is not immediately intuitive

In FRAME, pallet sudo is commonly used to guard functionality behind a signature by the `sudo` key. This is similar to hoe `SigCheck` works except that the key is dynamic. In Tuxedo this cannot be implemented as a redeemer because the sudo key is stored in state (a utxo) but redeemers do not have access to other state.

This could also be solved with evictions. (Again, see below).

### Con 3: Paying a private user

[...]

PR #41, opened 2023-03-07
Adding new simple verifier for decoupling

coax1d:
This is the start of the migration to a new simple verifier as discussed in #27 this includes the following tasks:
- [x] migrate old verifiers to new `SimpleVerifier` trait.
- [x] update `Executive` logic
- [x] Migrate old tests and old code depending old verifier logic


JoshOrndorff:
Overall, I totally agree with the direction of this PR. I opened #42 which makes the macro unnecessary.

coax1d:
@JoshOrndorff  let me know what you think if this is good for now. Thanks!!

tuxedo-template-runtime/src/kitties.rs
```diff
@@ -559,4 +543,144 @@ mod test {
     fn check_child_num_breedings_non_zero_fails() {
         // TODO
     }
+
+    #[test]
+    fn breed_wrong_input_type_fails() {
```
coax1d, line 548: Not sure why these tests are popping up when they already exist on main

tuxedo-template-runtime/src/kitties.rs
```diff
@@ -1,34 +1,18 @@
...
-//! In order to submit a valid transaction you must strutucture it as follows:
-//! 1.) Input must contain 1 mom and 1 dad
-//! 2.) Output must contain Mom, Dad, and newly created Child
-//! 3.) A child's DNA is calculated by:
-//!         BlakeTwo256::hash_of(MomDna, DadDna, MomCurrNumBreedings, DadCurrNumberBreedings)
-//!
-//! There are a only a finite amount of free breedings available before it starts to cost money
-//! to breed kitties.
-//!
-
```
JoshOrndorff, line 18: Is this deletion intentional?

tuxedo-core/src/verifier.rs
```diff
@@ -29,10 +33,54 @@ pub trait Verifier: Debug + Encode + Decode + Clone {
...
+            .collect();
+
+        // Call the simple verifier
+        SimpleVerifier::verify(self, &input_data, &output_data)
+    }
+}
+
+// A trivial verifier that verifies everything. Not practical. More for testing
+// and for the sake of making things compile before I get around to writing the
+// amoeba nd PoE verifiers
```
JoshOrndorff, line 81:
```suggestion
/// Utilities for writing verifier-related unit tests
```

tuxedo-template-runtime/src/lib.rs
```diff
@@ -295,26 +294,26 @@ impl From<runtime_upgrade::VerifierError> for OuterVerifierError {
...
         &self,
-        input_data: &[DynamicallyTypedData],
-        output_data: &[DynamicallyTypedData],
+        inputs: &[Output<R>],
+        outputs: &[Output<R>],
     ) -> Result<TransactionPriority, OuterVerifierError> {
         Ok(match self {
-            Self::Money(money) => money.verify(input_data, output_data)?,
-            Self::FreeKittyVerifier(free_breed) => free_breed.verify(input_data, output_data)?,
+            Self::Money(money) => Verifier::verify(money, inputs, outputs)?,
```
JoshOrndorff, line 303: I'm surprised this change is necessary. Of course we are using the new, more general, `Verifier` trait, but I didn't expect that we should have to explicitly mention it as long as the `SimpleVerifier` is not in scope.

JoshOrndorff (approved):
Looks basically good. I made a few suggestions, and we need the CI green. Then I'm fine with merging.

tuxedo-template-runtime/src/lib.rs
coax1d: You are correct I had initially the `SimpleVerifier` in scope. Am changing

tuxedo-template-runtime/src/kitties.rs
coax1d: Nope thx for catching I made this PR in a rush.

PR #42, opened 2023-03-08
Blanket impl for powerful verifier

JoshOrndorff:
:warning: This PR is not against `main`; it is against #41 

**Main Change**

The main thing this PR does is add a blanket implementation so that any type that implements the `SimpleVerifier` automatically implements the more powerful `Verifier`. Doing it with a blanket impl avoids the need for a macro.

**Secondary Change**

This slightly changes the signature for the new powerful `Verifier` trait so that it doesn't take inputs, but rather the `Outputs` that are being consumed by the inputs. This allows trait implementers (aka piece developers) not to worry about fetching the Outputs from storage.

### DIFF
--- a/tuxedo-core/src/verifier.rs
+++ b/tuxedo-core/src/verifier.rs
@@ -48,8 +48,7 @@ pub trait Verifier: Debug + Encode + Decode + Clone {
 // the executive can always just call the Powerful trait.
 impl<T: SimpleVerifier> Verifier for T {
 
-    // When using the blanket implementation, we directly return a UtxoError here
-    // So that we have access to the `RedeemerError` variant
+    // Use the same error type used in the simple implementation.
     type Error = <T as SimpleVerifier>::Error;
 
     fn verify<R: Redeemer>(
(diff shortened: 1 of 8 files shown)


tuxedo-core/src/verifier.rs
```diff
@@ -38,11 +38,43 @@ pub trait Verifier: Debug + Encode + Decode + Clone {
...
+
+        // Extract the output data
+        let output_data: Vec<DynamicallyTypedData> = outputs
+            .iter()
+            .map(|o| o.payload.clone())
+            .collect();
+
+        // Call the simple verifier
+        SimpleVerifier::verify(self, &input_data, &output_data)
+    }
```
JoshOrndorff, line 75: This is the main change. A simple addition that makes the macro unnecessary.

tuxedo-core/src/verifier.rs
```diff
@@ -38,11 +38,43 @@ pub trait Verifier: Debug + Encode + Decode + Clone {
     /// The actual verification logic
     fn verify<R: Redeemer>(
         &self,
-        inputs: &[Input],
+        inputs: &[Output<R>],
```
JoshOrndorff, line 41: This is the secondary change.

tuxedo-core/src/verifier.rs
coax1d: Ahh ok you can do this perfect!

coax1d (approved):
Thanks!

PR #45, opened 2023-03-13
Renaming to IOHK names

coax1d:
This addresses the last bit of #27 considering naming:

- [x] Verifier -> ConstraintChecker
- [x] Redeemer -> Verifier
- [x] Witness -> Redeemer


tuxedo-core/src/constraint_checker.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    /// The actual check validation logic
+    fn check(
+        &self,
+        input_data: &[DynamicallyTypedData],
+        output_data: &[DynamicallyTypedData],
+    ) -> Result<TransactionPriority, Self::Error>;
+}
+
+pub trait ConstraintChecker: Debug + Encode + Decode + Clone {
```
JoshOrndorff, line 32: We should have some docs on this trait. Maybe the ones form the old `SimpleConstraintChecker` trait should go here and those ones be updated.

tuxedo-core/src/utxo_set.rs
```diff
@@ -1,25 +1,23 @@
-//! These three functions comprize what I sketched out in the UtxoSet trait in element the other day
-//! For now, this is complicated enough, so I'll just leave them here. In the future it may be wise to
-//! abstract away the utxo set. Especially if we start doing zk stuff and need the nullifiers.
+//! TODO explanation this type
```
JoshOrndorff, line 1: :eyes:

JoshOrndorff (approved):
Overall good. Just a few little comments.

tuxedo-core/src/utxo_set.rs
coax1d: Ok ill leave it in wasnt sure what we should do here.

tuxedo-core/src/constraint_checker.rs
coax1d: Will do !

JoshOrndorff:
Looks good to me. Needs `cargo fmt` to make that CI pass.

The link checker CI will not pass on this one because you (correctly) include a link to the new trait's docs. Those docs will not be built and published until this is merged to `main` though. So no need to make that CI pass.

PR #49, opened 2023-03-17
Implement Peeks

JoshOrndorff:
This PR adds the peek functionality for reading state without consuming it as described in the original grant proposal.

This will help facilitate concurrency down the road. Even before that, it makes it nice that you can read state without re-writing it and therefore changing its `OutputRef`.


coax1d (approved):
Looks Solid nothing really to comment or give feedback here on. Nice!

JoshOrndorff:
TODO I believe the wallet still needs updated.

Issue #26, opened 2023-03-02
Multisig


JoshOrndorff:
For milestone 2 of the [original grant proposal](https://github.com/w3f/Grants-Program/blob/master/applications/tuxedo.md#milestone-2--wallet-and-multisig) we need to implement a multisig wallet. As described in Section 2 of the [EUTXO paper](https://jmchapman.io/papers/eutxo.pdf) there are two ways to do a multisig, and I propose we **implement both**.

**`Verifier`-Based Approach**

The simple way is to create a `Varifier` implementation that is similar to `SigCheck` but requires a certain number of valid signatures rather than just a single one. This kind of multi-sig has been around since bitcoin. It is also somewhat inconvenient for users because it requires aggregating all of the signatures through off-chain communication between the signatories.

```rust
/// A Threshold multisignature. Some number of member signatories collectively own inputs
/// guarded by this verifier. A valid redeemer must supply valid signatures by at least
/// `threshold` of the signatories. If the threshold is greater than the number of signatories
/// the input can never be consumed.
struct ThresholdMultiSignature {
  /// The minimum number of valid signatures needed to consume this input
  threshold: u8,
  /// All the member signatories, some (or all depending on the threshold) of whom must
  /// produce signatures over the transaction that will consume this input.
  signatories: Vec<H256>,
}
```

**`ConstraintChecker`-Based Approach**

The second way, as outlined in EUTXO, is to allow aggregating the signatures on-chain in a smart contract like way. This would be implemented as a `ConstraintChecker`. A signatory makes a proposal by consuming the original multisig and creating a new one with all the same data plus their proposal plus their own signature. Then other signatories would consume that utxo and replace it with one with all the same data plus their own signature. And eventually the final signatory would consume the utxo and replace it with just the proposal.

coax1d:
#50

PR #50, opened 2023-03-20
Adding multisig types with tests on new verifier

coax1d:
This adds a new verifier which handles multi-sigs as discussed in #26. In this issue there are two approaches and this PR addresses the first approach.

Checklist:
- [x] Add new `MultiSig Types`
- [x] Implement `Verifier Trait` on new `ThresholdMultiSignature` type
- [x] Add isolated `Verifier` tests
- [x] Add new `Verifier` to `OuterVerifier` in tuxedo-template-runtime


JoshOrndorff:
I sketched out an idea for how to ensure that no duplicate signatories exist at creation time. https://github.com/Off-Narrative-Labs/Tuxedo/pull/51 But the more I think about it, the more I like the way you handled it here. The only slight disadvantage is that a multisig with duplicate signatories will create a utxo that is forever stuck in state. But maybe this can be better solved by having some kind of cleanup transaction that allows anyone to remove an unspendable utxo for a reward.

coax1d:
I can build release with no issues and check and tests all pass is something weird going on with caching in the CI? If you look the code it is complaining about the missing `outerverifier` arm is already satisified.
https://github.com/Off-Narrative-Labs/Tuxedo/actions/runs/4473038238/jobs/7859921163?pr=50#step:6:435

JoshOrndorff:
Confirmed. `cargo test`, `cargo check`, and `cargo +nightly clippy` all pass locally for me as well. I've merged `main` into this branch to see if that helps. I don't see why it would, but worth checking. If CI caching slows us down, I'm very okay with getting rid of it.

PR #51, opened 2023-03-20
Verifier creation time checks

JoshOrndorff:
This is a braindump of an idea that allows verifiers to enforce constraints about their data at the time they are being created. I'm not sure if it is a good idea.

On the one hand, this prevents users from accidentally locking their funds or other assets forever by accidentally submitting invalid verifiers (eg a multisig with duplicated signatories).

On the other hand, the adds more checking logic on chain for all nodes to execute, and it only benefits the user who may have carelessly submitted an invalid verifier. When framed this way, I'm inclined to think that providing an invalid verifier is just a generalization on sending funds to the wrong account. In that case, it is culturally well established that this is the user's problem.

### DIFF
--- a/tuxedo-core/src/executive.rs
+++ b/tuxedo-core/src/executive.rs
@@ -79,6 +79,7 @@ impl<B: BlockT<Extrinsic = Transaction<V, C>>, V: Verifier, C: ConstraintChecker
         }
 
         // Make sure no outputs already exist in storage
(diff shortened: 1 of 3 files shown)


JoshOrndorff:
Closing this because we liked #50 better without it. I'm glad I wrote it down in case we need it in the future, but right now I'm happy without it.

Issue #16, opened 2023-03-02
Peeks


JoshOrndorff:
In the [original grant proposal](https://github.com/w3f/Grants-Program/blob/master/applications/tuxedo.md) I sketched a transaction type that included `peeks` alongside traditional UTXO inputs.

```rust
/// A UTXO transaction specifies some inputs to be consumed, and some new outputs to be created.
struct Transaction {
  /// The inputs refer to currently existing unspent outputs that will be consumed by this transaction
  inputs: Vec<Input>,
  /// Similar to inputs, Peeks refer to currently existing utxos, but they will be read only, and not consumed
  peeks: Vec<OutputRef>,
  /// The new outputs to be created by this transaction.
  outputs: Vec<Output>,
}
```

When writing that, I conceived that peeks would help with concurrency for transaction that needed to read state from existing UTXOs but didn't need to modify it and therefore would prefer not to consume it. Having worked on the PoC, we now have a [concrete use case](https://github.com/Off-Narrative-Labs/Tuxedo/blob/7f1dabe37ce8621deb9df47d102ea775bbe2f763/frameless-runtime/src/poe.rs#L144-L147) for peeks as well.

muraca:
Is that #49 ?

PR #52, opened 2023-03-21
Implement Evictions

JoshOrndorff:
This PR builds off of #49. And it currently uses that branch as its base branch to make the eviction-related changes more obvious.

This PR is open for discussion and review. I'm not sure whether this is the best way to proceed, but it is an idea I've had around for a while so I decided to code it up.

In general, evictions are a way for a tracnsaction to read and consume some state (similar to an input) without having to satisfy the input's verifier. This could be used to clean up old or irrelevant state. For one example, it could be used to clean up `ThresholdMultiSignature`s that were constructed with duplicate signatories and are therefore never able to be consumed (It is an alternative to #51 in that sense).

For another example, see the `PoeDispute` `ConstraintChecker` included in this PR. This allows owners of proof-of-existence claims to evict claims for the same data that came later.

### DIFF
--- a/tuxedo-core/src/constraint_checker.rs
+++ b/tuxedo-core/src/constraint_checker.rs
@@ -26,6 +26,7 @@ pub trait SimpleConstraintChecker: Debug + Encode + Decode + Clone {
         &self,
         input_data: &[DynamicallyTypedData],
         peek_data: &[DynamicallyTypedData],
+        eviction_data: &[DynamicallyTypedData],
         output_data: &[DynamicallyTypedData],
     ) -> Result<TransactionPriority, Self::Error>;
 }
@@ -48,6 +49,7 @@ pub trait ConstraintChecker: Debug + Encode + Decode + Clone {
         &self,
         inputs: &[Output<V>],
         peeks: &[Output<V>],
+        evictions: &[Output<V>],
         outputs: &[Output<V>],
     ) -> Result<TransactionPriority, Self::Error>;
 }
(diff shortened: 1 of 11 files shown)


JoshOrndorff:
This PR along with #16 have been open for a long time and are stale. That is because I was thinking about the right design, and I believe I have finally reached a conclusion.

Peeks will go in basically as they are in #16 once it is resolved. Evictions will be redesigned and this PR closed.

Some rationale:

In the UTXO model an offchain tool can run the chain forwards and backwards calculating state by only looking at the blocks themselves, not the runtime logic. The transaction itself tells exactly which UTXOs must pre-exist, which of those will be consumed, and which new ones will be created. This clean and simple property should be preserved in its simplest form.

In the case of Peeks it is necessary to treat them as an equal with an input or output because they require UTXOs to exist. Evictions OTOH are not so fundamental. An eviction is just a bit of runtime logic that says it is okay to use a UTXO without satisfying the verifier, and wallets do not need to know this level of detail.

I propose that the correct way to design what I previously thought of as "evictions" is as follows. Stop unconditionally checking verifiers in the core transaction execution flow. Generalize the notion of `ConstraintChecker` as we did in #41 This time splitting it into 3 levels.

1. The simple one we've always had. Checks verifiers for you, just gives you data
2. The one added in #41. Checks verifiers for you, gives you both data and verifiers.
3. The new one which is the most powerful of all. It does not check verifiers for you and gives you both data and verifiers.

If your operation will be an eviction, it will need to implement the new 3rd kind of constraint checker. And if it wants to check _some_ of the verifiers, it will need to do so manually.

PR #60, opened 2023-04-07
Filter based on keystore

JoshOrndorff:
The primary change of this PR is that the wallet decides which outputs are relevant based on the keys in the keystore instead of the fixed SHAWN key.

While making this change, I noticed that the filter can be passed directly without the need for the builder. @coax1d Please lmk if I have missed something here. The signature you wrote for the build function got me thinking about how to pass functions around, and I think we can do it without needing a builder. Although if I've misunderstood, please let me know.

### DIFF
--- a/wallet/src/main.rs
+++ b/wallet/src/main.rs
@@ -7,7 +7,6 @@ use jsonrpsee::{
     http_client::{HttpClient, HttpClientBuilder},
     rpc_params,
 };
-use keystore::SHAWN_PHRASE;
 use parity_scale_codec::{Decode, Encode};
 use runtime::OuterVerifier;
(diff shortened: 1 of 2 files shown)


coax1d (commented):
I dont mind doing it this way I was just making the build function because we can extend the building of filters in its own custom module. This also can work although the F type is now bound to a specific closure signature instead of a more generic filter type. I was hard coding the Filter type but that was just a temporary solution. Just thinking ahead to filtering by utxo type verifier etc

coax1d:
Still feel there can be some value having many different ways to build filters for utxos.

coax1d:
Looks good!

PR #63, opened 2023-04-11
Wallet Readme

JoshOrndorff:
Aka Joshy's final pass before milestone 2.

This PR:
* Adds a readme with a detailed walkthough in the wallet crate
* Fixes a bug where the wallet would select the same input multiple times for inclusion in a transaction
* Uses `env_logger` and `log` crates for better output logging
* Makes the CLI a little more intuitive

### DIFF
--- a/wallet/src/sync.rs
+++ b/wallet/src/sync.rs
@@ -225,10 +225,13 @@ pub(crate) fn get_unspent(db: &Db, output_ref: &OutputRef) -> anyhow::Result<Opt
 
 /// Picks an arbitrary set of unspent outputs from the database for spending.
 /// The set's token values must add up to at least the specified target value.
-/// 
+///
 /// The return value is None if the total value of the database is less than the target
 /// It is Some(Vec![...]) when it is possible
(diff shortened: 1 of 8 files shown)


wallet/src/sync.rs
```diff
@@ -224,17 +223,35 @@ pub(crate) fn get_unspent(db: &Db, output_ref: &OutputRef) -> anyhow::Result<Opt
...
-    };
 
-    let output_ref = OutputRef::decode(&mut &output_ref_ivec[..])?;
-    let (owner_pubkey, amount) = <(H256, u128)>::decode(&mut &owner_amount_ivec[..])?;
+    let mut total = 0u128;
+    let mut keepers = Vec::new();
+
+    let mut unspent_iter = wallet_unspent_tree.iter();
+    while total < target {
+        let Some(pair) = unspent_iter.next() else {
```
coax1d, line 242: This is cool didnt know you could do this

coax1d (approved):
Looks good. Very concise instructions I like that this is geared towards the person who is only vaguely familiar with UTXOS. I think we will have some uphill battle helping people understand UTXO model in general. Sweet!

PR #65, opened 2023-04-12
Make `CostraintChecker` trait generic over `Verifier`

JoshOrndorff:
This PR changes the Constraint Checker trait to have a generic type `Verifier` as opposed to its method having the generic. More details in the comments below.

This PR originally began as my sketch for an order book dex. But it ended up being just the core changes I needed to support such a dex. The dex itself is no longer included in this PR.

**Original Description**:

> This PR adds a new tuxedo piece which is a simple order-book based decentralized exchange between two tokens.
> 
> I started sketching this out in hopes that we might be able to show it at the Substrate Seminar.


coax1d:
Overall this looks good. In the future we should wrap up the coupling into a standardized generic way of doing coupling such as a Config trait. Great job getting this working.

JoshOrndorff:
The main thing that I dislike about this PR is that it touches all the way down to the aggregation macros. And it makes them more complicated.

That happened because I moved a type parameter from a method to the `ConstraintChecker` trait as shown here.
```rust
// Before
pub trait ConstraintChecker {
    fn check<V: Verifier>(
        &self,
        inputs: &[Output<V>],
        outputs: &[Output<V>],
    ) -> Result<TransactionPriority, Self::Error>;
}

// After
pub trait ConstraintChecker<V: Verifier> {
    fn check(
        &self,
        inputs: &[Output<V>],
        outputs: &[Output<V>],
    ) -> Result<TransactionPriority, Self::Error>;
}
```

The reason I encountered this is because I want to make sure that the `payload_verifier: Verifier` stored in the Order is the same concrete type as the Verifier that the Constraint Checker is able to use in inputs and outputs.

After reading https://stackoverflow.com/a/53085395/4184410 I'm okay with it, because I think it really does make more sense to have the verifier be a generic type on the trait because that allows the implementer to further constrain it.

In this case I'm expressing, "The `MatchOrders` type only implements `ConstraintChecker` when the `ConstraintChecker`'s verifier type is the same as the verifier type stored in the `Order`s". Such a constraint is not possible when the generic is on the method, because the method's generics are only constrained by the trait definition itself.

Wow, finally convinced myself.

JoshOrndorff:
So what I actually want to do with this PR is remove the dex itself, and just merge the changes to Tuxedo. The big ones are the instantiable money piece and the constraint checker trait change.

Then I'll put the dex and the corresponding tutorial in its own repo.

Issue #14, opened 2023-03-02
Aggregation Macros


JoshOrndorff:
Writing a complete Tuxedo runtime will typically involve composing several individual pieces together. In the current design this means creating aggregate verifier and redeemer enums that have a variant for each of the individual verifiers and redeemer. Currently there is no macro to make this job easier, and all of the code must be written manually, which is long and redundant.

**Simple Aggregation**

At minimum we will want a macro to provide all of the `From` implementations. I've prototyped such a macro https://github.com/JoshOrndorff/proc-macro-experiment/blob/main/src/lib.rs. This is a good start. In fact this same macro could be used to aggregate the verifier and the redeemer (and any other types where we want this pattern in Tuxedo or elsewhere).

**Implementing `Verifier` on the aggregate enum**
The simple aggregator I've sketched is a good start, however we may want it to do more. For example, we may want it to implement the `Verifier` trait on the aggregate verifier enum. This implementation is mechanical because each of the inner types are already `Verifier` so we just have to match on the variant and then dispatch to the inner type. This would require dedicated macros for the verifier and the redeemer because the traits have different methods with different signatures.

**Syntax**
Assuming we want the traits implemented by the macros, there is the question of syntax. I can imagine two possibilities and there may well be more.

Option 1: Single macro with parameters used as follows
```rust
#[aggregate(Verifier)]
pub enum AggregateVerifier {
  Money(money::Verifier),
  Amoeba(amoeba::Verifier),
}

#[aggregate(ConstraintChecker)]
pub enum AggregateChecker {
  SigCheck(Sigcheck),
  UpForGrabs(UpForGrabs),
}
```

Option 2: Separate macros for the verifier and the redeemer
```rust
#[aggregate_verifier]
pub enum AggregateVerifier {
  Money(money::Verifier),
  Amoeba(amoeba::Verifier),
}

#[aggregate_redeemer]
pub enum AggregateRedeemer {
  SigCheck(Sigcheck),
  UpForGrabs(UpForGrabs),
}
```

In the second case, both macros may call out to a third macro that is similar to my prototyped helper for the aggregation alone. Currently I'm leaning toward option 1.

**Additional Design Considerations**

It is possible that the scope of the macros' jobs may still change. So even when we decide something here, there is no guarantee that we will stick with it forever and ever amen.

JoshOrndorff:
[...]

PR #66, opened 2023-04-13
Aggregation macros

JoshOrndorff:
Solves #14 

This PR introduced the `aggregate` macro with automates the process of aggregating several tuxedo pieces into a single runtime.

- [x] Basic outer enum aggregation
- [x] Auto derive Verifer
- [x] Auto derive ConstraintChecker
- [x] Autodefine constraint checker error enum

I'm doing this now because I think having the macros will make Tuxedo more familiar and approachable to learners who encounter it on the upcoming Substrate Seminar.

### DIFF
--- a/tuxedo-core/aggregator/src/lib.rs
+++ b/tuxedo-core/aggregator/src/lib.rs
@@ -3,10 +3,10 @@ use quote::quote;
 use syn::{parse_macro_input, Ident, ItemEnum};
 
 /// Automatically implements `From` for each type in an aggregate type enum.
-/// 
+///
 /// The supplied enum should have a single unnamed type parameter for each variant.
 /// And the type for each variant should be unique in the enum.
-/// 
+///
 /// The macro generates all the `From` implementations automatically.
 #[proc_macro_attribute]
 pub fn aggregate(_: TokenStream, body: TokenStream) -> TokenStream {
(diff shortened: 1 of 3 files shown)


tuxedo-core/aggregator/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+    });
+    let variants = variant_type_pairs.clone().map(|(v, _t)| v);
+    let inner_types = variant_type_pairs.map(|(_v, t)| t);
+
+    // The macro supports automatically implementing Tuxedo-related traits.
+    // Currently two traits (Verifier and ConstraintChecker) are supported,
+    // as well as no trait auto-implementation. More traits may be supported in the future.
+    let mut tuxedo_trait_impl = quote! {};
+
+    if !attrs.is_empty() {
```
coax1d, line 39: Think would be best to separate out each of verifier and constraint checker into their own macros to avoid this growing very very large

coax1d (commented):
Minor comments overall I like reducing the boilerplate. Is there anyway to make any isolated unit tests? this isnt totally necessary in my opinion right now but more of a curiosity and question for the future of proc macros

tuxedo-core/aggregator/src/lib.rs
JoshOrndorff: Agreed. I've done it now.

PR #67, opened 2023-04-17
initial attempt - just search and replace

JoshOrndorff:


tuxedo-core/src/verifier.rs
```diff
@@ -164,14 +164,30 @@ mod test {
     use super::*;
     use sp_core::{crypto::Pair as _, sr25519::Pair};
 
+    /// Generate a bunch of test keyparis
```
coax1d, line 167:
```suggestion
    /// Generate a bunch of test keypairs
```

coax1d (approved):
Looks good is this all that was required for updating?

JoshOrndorff:
No, there is more needed. The client-side doesn't compile yet.

JoshOrndorff:
This is almost done now. Looks like there is one legit test failure:

```
test verifier::test::threshold_multisig_not_enough_sigs_fails ... FAILED
```

I reproduced it locally. Haven't dug in yet.

PR #69, opened 2023-04-19
Multitoken

JoshOrndorff:
This PR explores allowing multiple tokens in a single runtime... twice. That is, it explores two different designs for a multitoken tuxedo piece.

## Strongly Typed Design

This first design is included as a change to the existing money piece. It just adds a `<const ID: u8>` generic to the `Coin` type. This allows different tokens to be distinguished _at the type level_. You can never send the wrong kind of token because the type system won't allow it.

This design is nice for use in our upcoming DEX tutorial. I plan to make the DEX generic over the types of the two tokens in the trading pair. This design gives a dedicated type to each token which will work nicely.

One consequence (which my be desirable or not depending on application) is that the number of tokens in the runtime must be fixed at compile time. You must explicitly add the `MoneyConstraintChecker::<2>` to your outer checker for each and every token you plan to use.

## Dynamically Typed design

The second design is in the new multitoken piece. This is still based on the original money code, but takes a much different approach. It replaces the Coin struct with
```rust
pub struct Coin {
  pub id: u8,
  pub value: u128,
}
```

This allows tokens to be added and destroyed from the runtime at any moment and no fixed number is required.

However, there is still only a single type for all the tokens. This means that if we want to used this with our DEX, we will need to make run-time checks on the token id all over the place.

## Are both useful?

Perhaps both should be kept in the library as they may have slightly different use cases.

### DIFF
--- a/tuxedo-template-runtime/src/lib.rs
+++ b/tuxedo-template-runtime/src/lib.rs
@@ -30,6 +30,7 @@ use serde::{Deserialize, Serialize};
 pub mod amoeba;
 pub mod kitties;
 pub mod money;
+pub mod multitoken;
 mod poe;
 mod runtime_upgrade;
 use tuxedo_core::{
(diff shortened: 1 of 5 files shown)


coax1d (commented):
I think I reviewed these in the wrong order. I like the Strongly Typed approach best. Creating an issue for it #75

JoshOrndorff:
I also like the strongly typed version better. I've cherry-picked it over to #65 already, so I'll close this, and let the multitoken stuff go in with the dex.

PR #71, opened 2023-04-19
Update dockerfile to match Substrate's

JoshOrndorff:
Docker CI normally only runs on `main`. When I merged the aggregation macros PR, it failed on main. I reran a few times over the next few days and it is still failing. IDK why. The log output sounds like the wasm32-unknown-unknown target isn't installed, but it is running in a parity builder image. It is still recommended in the official Substrate repo.

Although the docs around that docker setup have changed a bit. They got removed from the node template repo, but still exist in the main Substrate repo. IDK if that is related to our now-failing CI. I've updated the links in any case.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -10,7 +10,6 @@ on:
   push:
     branches:
       - main
-      - docker-ci-fix-hopefully
     tags:
       - '**'
 
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -10,6 +10,7 @@ on:
   push:
     branches:
       - main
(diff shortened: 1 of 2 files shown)


JoshOrndorff:
Well, the CI seems to working again on `main` on its own. I guess there was some problem with the Parity builder image, but I couldn't find any docs or PR or anything about it.

Anyway, this PR still makes a few little updates to bring our dockerfile in line with Substrate's (linked in the dockerfile itself) and also [Polkadot's](https://github.com/paritytech/polkadot/blob/master/scripts/ci/dockerfiles/polkadot/polkadot_builder.Dockerfile)

PR #72, opened 2023-04-20
Another multisig test

JoshOrndorff:
When I was recently preparing the Substrate update in #67, I was briefly confused about the tests on the `ThresholdMultisig`. Once I read the tests carefully, I understood them and quickly got them passing.

However, I noticed that there was not a test for the case where the redeemer contains _more than enough_ valid signatures. This PR adds a single test for when there is a 2/3 multisig, and the redeemer contains valid sigs for _all three_ signatories even though only two are required. I expected this case to still successfully verify, but it seems it is not.

@coax1d is this the behavior you expected?

### DIFF
--- a/tuxedo-core/src/verifier.rs
+++ b/tuxedo-core/src/verifier.rs
@@ -253,6 +253,32 @@ mod test {
     }
 
     #[test]
+    fn threshold_multisig_extra_sigs_still_passes() {
+        let threshold = 2;
+        let pairs = generate_n_pairs(threshold + 1);
+
+        let signatories: Vec<H256> = pairs.iter().map(|p| H256::from(p.public())).collect();
+
+        let simplified_tx = b"hello_world".as_slice();
+        let sigs: Vec<_> = pairs
+            .iter()
+            .enumerate()
+            .map(|(i, p)| SignatureAndIndex {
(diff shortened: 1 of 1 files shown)


tuxedo-core/src/verifier.rs
```diff
@@ -252,6 +252,32 @@ mod test {
...
+            })
+            .collect();
+
+        let redeemer: &[u8] = &sigs.encode()[..];
+        let threshold_multisig = ThresholdMultiSignature {
+            threshold,
+            signatories,
+        };
+
+        assert!(!threshold_multisig.verify(simplified_tx, redeemer));
```
coax1d, line 278:
```suggestion
        assert!(threshold_multisig.verify(simplified_tx, redeemer));
```

This will make it pass

coax1d (commented):
Think there was just a small glitch if I understand correctly

tuxedo-core/src/verifier.rs
JoshOrndorff: Lol. Thanks for checking. :facepalm:

coax1d:
Looks good tho good test to add!

coax1d (approved):
LGTM!

PR #73, opened 2023-04-21
Verifier Abstractions

JoshOrndorff:
Inspired by @nukemandan and @franciscoaguirre.

This PR introduces (more like sketches out, for now) a Tuxedo Piece that provides a way to dynamically swap verifiers for an unbounded number of utxos in a single transaction.

This is basically the UTXO equivalent of account abstractions.


JoshOrndorff:
Closing this because it is stale. It may also be better done with something like #186 .

I'm open toexploring this further in the future if there is desire.

Issue #22, opened 2023-03-02
Opaque Extrinsic


JoshOrndorff:
We don't currently have an opaque extrinsic type (I think). Honestly this part was never very clear to me. I think we need a separate opaque extrinsic type to support runtime upgrades. Currently there is still a [comment from the frameless runtime template](https://github.com/Off-Narrative-Labs/Tuxedo/blob/7f1dabe37ce8621deb9df47d102ea775bbe2f763/frameless-runtime/src/lib.rs#L51-L55)

```rust
pub mod opaque {
    use super::*;
    // TODO: eventually you will have to change this.
    type OpaqueExtrinsic = Transaction;
    // type OpaqueExtrinsic = Vec<u8>;

    // --snip--
}
```

JoshOrndorff:
Indeed, this was closed in #74

PR #74, opened 2023-04-21
Opaque Extrinsics

JoshOrndorff:
Solves #22 

I set out just trying to make the derives in the runtime look a little nicer. Ended up here. I do think they look nicer though, so mission accomplished.

This PR changes the encoding of the `tuxedo_core::Transaction<V, C>` slightly. It prepends the previous derived encoding with the length of that encoding in bytes. The whole point of this is that runtime's can then use the `OpaqueExtrinsic` type so the client doesn't need to know the details and runtime upgrades are more useful.

The actual encoding and decoding logic is inspired by @coax1d's PBA example solution https://github.com/Polkadot-Blockchain-Academy/frameless-node-template--master/blob/andrew_ungraded_solution/frameless-runtime/src/lib.rs#L127-L144. But I'm not so sure the decoding is right over there. I don't see it skipping the prepended compact length anywhere. I added that here and it seems to work fine.

### DIFF
--- a/tuxedo-core/src/dynamic_typing.rs
+++ b/tuxedo-core/src/dynamic_typing.rs
@@ -53,10 +53,7 @@ use sp_std::vec::Vec;
 
 /// A piece of encoded data with a type id associated
 /// Strongly typed data can be extracted
-#[cfg_attr(
-    feature = "std",
-    derive(Serialize, Deserialize, parity_util_mem::MallocSizeOf)
-)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
 #[derive(Encode, Decode, Debug, PartialEq, Eq, Clone)]
 pub struct DynamicallyTypedData {
     pub data: Vec<u8>,
--- a/tuxedo-core/src/types.rs
+++ b/tuxedo-core/src/types.rs
@@ -35,10 +35,7 @@ pub struct OutputRef {
 ///
 /// In the future, there may be additional notions of peeks (inputs that are not consumed)
 /// and evictions (inputs that are forcefully consumed.)
-#[cfg_attr(
-    feature = "std",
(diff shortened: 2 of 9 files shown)


tuxedo-core/src/executive.rs
```diff
@@ -898,7 +898,7 @@ mod tests {
                         "858174d563f845dbb4959ea64816bd8409e48cc7e65db8aa455bc98d61d24071",
                     ),
                     extrinsics_root: array_bytes::hex_n_into_unchecked(
-                        "7ceffb73687cb9af3ad2f9a0c544a216df70894b03da3ceb57ead37bd6b51be0",
+                        "b3dd85732bb51b79d7c5c3518cfb1b202210743b9dc7b8ccf848f9a66e0a0f6b",
```
coax1d, line 901: What is this change about ? Just curious.

coax1d (approved):
Overall looks good and a useful change.

tuxedo-core/src/executive.rs
JoshOrndorff: The extrinsics root is the merkle tree root that you get when you build all of the extrinsics into a merkle tree. Since the extrinsics encoding changed here (to have the length prepended) the root also changed.

Notice it only changed in this test where there is actually an extrinsic in the block. The other tests where the block is empty still have the same root as before. This is good and confirms this test only changed because of the new encoding, not some other weird bug I introduced.

I didn't manually calculate that new root or expect that exact hash. I just noticed that it changed (because the test failed) decided it made sense that it changed (as explained above) and copied the new root. AFAICT that's what the FRAME tests do too. That's what I did to get the roots when I first wrote those tests anyway.

This does feel a tiny bit questionable, but I rationalized it saying the test still adds value because it lets us know any time the value changes. And if it changes when we don't expect it to, that means we have a problem. If it changes when we do expect it to (like now) we can just update the value.

PR #80, opened 2023-05-03
Full dex solution

JoshOrndorff:
DO NOT MERGE

This PR is to show the diff of the dex against what is already on master.


coax1d (approved):
The only part of the Dex that I think is sorta confusing is the ReverseConfig part. Other than that I really like the Config aspect. Granted the number of types can get Large and is not really that easily changed.(I.E. if you want to add a new token to your dex or set of tokens then you really have to do a runtime upgrade)

All of that being said the config itself really allows the user of Tuxedo to get feel some real nice familiarity with FRAME. 

Just when thinking about this though, is there some more dynamic way of adding and removing certain types of coins for swaps? All in all this is very concise imo and a good demonstration of what we can do initially with Tuxedo.

JoshOrndorff:
Closing because this content is now a proper tutorial at https://github.com/Off-Narrative-Labs/Tuxedo-Order-Book-Dex-Tutorial/

PR #91, opened 2023-09-22
Use Workspace Dependencies

### LINKED ISSUE #85: Use Workspace Dependencies
A lot of Substrate projects are now specifying dependencies in the main `Cargo.toml`.

Tuxedo should do that too.

muraca:
closes #85

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,3 +17,71 @@ members = [
 
 [profile.release]
 panic = "unwind"
+
+[workspace.dependencies]
+# generic dependencies
+log = "0.4"
+serde =  "1.0"
+parity-scale-codec = { version = "3.4.0", default-features = false }
+scale-info = { version = "2.1.1", default-features = false }
+parity-util-mem = "0.12.0"
+jsonrpsee = "0.16.2"
+hex-literal = "0.4.1"
+clap = "4.3.0"
+
+# core-only dependencies
+array-bytes = "6.0.0"
+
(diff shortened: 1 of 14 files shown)


Cargo.toml
```diff
@@ -17,3 +17,76 @@ members = [
...
+quote = "1.0.23"
+proc-macro2 = "1.0.67"
+
+# core-only dependencies
+array-bytes = "6.0.0"
+
+# wallet-only dependencies
+serde_json = "1.0"
+tokio = "1.25.0"
+anyhow = "1.0.69"
```
JoshOrndorff, line 43: I like this separation of the dependencies in many ways. I also like having them alphabetical so that we can easily have a CI keep it in order.

For now let's leave it as it is except capitalize each line like this
```suggestion
# Wallet-only dependencies
serde_json = "1.0"
tokio = "1.25.0"
anyhow = "1.0.69"
```

I remember when I was at moonbeam, they wrote a tool called toml sort https://github.com/moonsong-labs/toml_sort for this exact purpose. I recommend that we open a follow up issue to add toml-sort to our ci.

@nanocryk are you guys still using / supporting toml-sort? I see no recent commits, but I guess that's just because it is simple and stable, right?

JoshOrndorff (approved):
Looks good to me assuming ci passes

Cargo.toml
muraca: I love this!

PR #94, opened 2023-09-26
Joshy update August 2023

JoshOrndorff:
Another routine dependency update. I'm currently stuck because `sp-statement-store` won't compile. Perhaps related to https://github.com/paritytech/substrate/pull/14776?


JoshOrndorff:
I'm not sure what's wrong with this one, but #95 is going better, and is a newer dependency anyway, so I'm gonna go straight to that one.

PR #100, opened 2023-09-27
Timestamp Inherent

### LINKED ISSUE #90: Timestamp Inherent
As we begin to prepare for parachain support, I want to master handling inherents first. Much of the parachain logic is in handling the parachains inherent, so we should be sure we have the fundamentals right. Plus many chains will find a timestamp piece useful; just look how ubiquitous pallet timestamp is in the accounts worlds.

My understanding is that we will be able to rely entirely on the existing client-side timestamp inherent worker which is already installed on the tuxedo template's client side ([importing](https://github.com/Off-Narrative-Labs/Tuxedo/blob/18c2ab36dcd4854baa6bf31aa76cf9f0b24b2acb/node/src/service.rs#L110), and [authoring](https://github.com/Off-Narrative-Labs/Tuxedo/blob/18c2ab36dcd4854baa6bf31aa76cf9f0b24b2acb/node/src/service.rs#L249)). This machinery is currently vestigial, but harmless as the runtime just ignores any provided inherent data at the moment.

Since we already have the inherent data coming in, we will just need to do some massaging (possibly including decoding or encoding) and then ultimately return a transaction that takes the corresponding action. In our case it will need to be a utxo style transaction.

Then the client side block authoring machinery inserts those inherents somewhere. The default implementation is to put them first which is perfect for timestamp inherent. But as a side note, I hope to get creative about that when tackling block author rewards based on transaction fees properly.

I can imagine two strategies for how the runtime might expect the transaction to handle the timestamp data
1. Very UTXO - Each inherent consumes the previous timestamp and produces a new one. Each tx must have exactly one input and one output. There would necessarily need to be a genesis timestamp then. One drawback here is that user transactions that need to peek at the timestamp (and therefore specify it by utxo id) would have a fundamental lifetime expiry on their transactions. This is unfortunate because you usually need to prove that it is later than x time, not immediately after x moment.
1a. UTXO - Same as above except that you don't clean up the previous one. Maybe there is an economic spam prevention mechanism like costing money to add a new timestamp and giving a reward for cleaning old ones up. Actually the economics could have a curve. Like the more you're updating the time by, the bigger the reward. This disincentivizes making small increment updates. same about rewarding more for cleaning up older ones. It might even cost you to clean up recent ones. If we are not cleaning up the UTXOs immediately, we may need a special one that points the most recent one. Then to update you really are racing for that tag.
2. Stateful - The tx has no inputs or outputs, it just has attached data which is the timestamp. That timestamp is then stored in a well-known location. This is very accounts. Not preferred.

I'll probably start by implementing option 1 because it is utxo friendly and simple. Maybe I'll do 1a if I'm on a roll, because I think it sounds better designed overall. Otherwise I'll make a followup issue about it.

One thing I'm not yet sure about is the notion of mandatory inherents. Is that a FRAME concept? If so we'll need to decide whether we duplicate it or not. I prefer simplicity, but if it is necessary, then it is necessary. I think it would be fine to have a block where the clock didn't tick because updating it wasn't economically feasible. OTOH, you may argue that keeping things like the clock or the parachain related info up to date is a top priority public good and should have dedicated blockspace.

JoshOrndorff:
Closes #90 

This main purpose of this PR is to blaze a path for how inherents can be used in Tuxedo.

As a concrete starting point we introduce the timestamp inherent which is ubiquitous in FRAME runtimes.

- [x] Minimal Tuxedo piece to accept inherent extrinsic
- [x] Install Piece in Runtime
- [x] Convert inherent data to extrinsic
- [x] Scrape parent block for input utxo
- [x] Insert extrinsic into block
- [x] Check inherent on importing node
- [x] Voluntary cleanup of old timestamps
- [x] Tests for the piece
- [x] `SetTimestamp` forbidden in pool (so it must be inherent)
- [x] Modularized design for arbitrary future inherents (eg parachain)
- [x] Aggregation macro
- [x] Move to creation and checking logic from template to executive
- [x] Enforce that inherents are first when executing
- [x] Tests for inherent ordering
- [x] Move configuration constants into the piece's configuration
- [ ] (Deferred to followup) Rework relationship between constraint checkers and inherent hooks.
- [x] (<del>Deferred until #105 lands</del>) Ensure exactly one time stamp per block.
- [ ] (Deferred until #107 is implemented) Remove the special case for initializing timestamp in the first block.


JoshOrndorff:
I'm starting to feel pretty good about this, but I'm running into a strange problem in a multinode network.

When I start a node with one author and one non-authoring full node, the full node imports the first block successfully, but then it is dropped as a peer. Looking at the authoring node's log, we see `Report 12D3KooWFVfCrqJx8eSWQdPoVDk2H5h2WCy8iNuyboR7zmVi7ZMc: -2147483648 to -2147483648. Reason: Same block request multiple times. Banned, disconnecting.` So the authoring node is upset that the peer is requesting the same block repeatedly, and drops the peer. This behavior makes sense. But what block is the peer requesting, and why over and over?

I checked the `main` branch and confirmed that multiple node networks are working as expected there, so somehow this PR is making the problem.

To debug, I tried manually exporting blocks from the authoring node to a json file via the cli, and them manually importing from the json file into the peer node via its cli as follows. This works! So the inherent checking seems correct.

```sh
You can export one block at a time like this
./target/release/node-template export-blocks --from 1 --to 1 -d /tmp/alice > block1.json
./target/release/node-template export-blocks --from 2 --to 2 -d /tmp/alice > block2.json

./target/release/node-template import-blocks ./block1.json -d /tmp/peer/
./target/release/node-template import-blocks ./block2.json -d /tmp/peer/

Or all at once like this
./target/release/node-template export-blocks -d /tmp/alice > all-blocks.json
./target/release/node-template import-blocks ./all-blocks.json -d /tmp/peer/
```

JoshOrndorff:
Adding `-lsync=debug` to both nodes, I can see that the is block two that is requested over and over, and the author handles it politely twice before unpeering.

The peer node that is trying to import the block also now tells us this:
`Bad extrinsic root for a block 0xfee6…71f6 received from 12D3KooWEYV8ojGAs9eJ3jVJZ1Fb9vpsPZBvrTStWRyFocYjm9Ao. Expected 0xa757f0a634ad7e06fdae892307ec496b76b98f205084f96b6e35dc49836fde22, got 0x89b9c4d9b59056f2dfd7af352f15065f400eb27f4d897d30a092e339f25c23c1  `

IDK why it goes on the request the block again after that, but I think this is our underlying problem. A Stateroot mismatch.

JoshOrndorff:
Oh wow, it looks like the Network syncing module is still using `StateVersion::V0`. I bet this is the problem. Why is that hard-coded?

JoshOrndorff:
Note to self and anyone else following along. This is really close. I'm currently debugging and issue where the second block fails to import because the input it tries to consume is apparently missing.

wardrobe/poe/src/lib.rs
```diff
@@ -85,6 +85,7 @@ pub trait PoeConfig {
 /// waste of caller fees.
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
 #[derive(Encode, Decode, DebugNoBound, CloneNoBound, PartialEq, Eq, TypeInfo)]
+#[scale_info(skip_type_params(T))]
```
JoshOrndorff, line 88: So what about this change? At some point this helped me beat a compiler error. But now that I've cleaned up the types and traits, it is no longer necessary to build. But I guess maybe it is a good idea anyway?

JoshOrndorff:
Looks like the CI is catching a real problem. The wallet isn't able to sync block one because it can't get the block from the node. I have no idea how that's relevant to this PR. But I confirmed it is not a problem on `main`.

Any help in diagnosing this is very very welcome.

tuxedo-core/src/executive.rs
```diff
@@ -360,14 +389,85 @@ impl<
...
+    pub fn inherent_extrinsics(data: sp_inherents::InherentData) -> Vec<<B as BlockT>::Extrinsic> {
+        debug!(
+            target: LOG_TARGET,
+            "Entering `inherent_extrinsics`."
+        );
+
+        // Extract the complete parent block from the inheret data
+        let parent: B = data
+            .get_data(&PARENT_INHERENT_IDENTIFIER)
+            .expect("1")
```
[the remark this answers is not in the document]
JoshOrndorff: Yeah, these were shitty messages. I've improved them in 057cdb7599ef3c8893774384ce340154dd9b141f.

tuxedo-core/src/executive.rs
```diff
@@ -291,8 +303,25 @@ impl<
...
+            // Enforce that inherents are in the right place
+            match (
+                extrinsic.checker.is_inherent(),
+                finished_with_opening_inherents,
+            ) {
+                (true, false) => (),                                      // Opening inherent
+                (false, false) => finished_with_opening_inherents = true, // First non-inherent
+                (false, true) => (),                                      // Other non-inherent
+                (true, true) => {
+                    panic!("Tried to execute opening inherent after switching to non-inherents.")
```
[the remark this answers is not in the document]
JoshOrndorff: Yes, this is the main `execute_block` function, arguably the most fundamental runtime API in Substrate. It does not return a Result. Panicing is the only correct thing to do when you are executing a block and find it to be invalid.

You can see FRAME also document's a panic in that function https://paritytech.github.io/polkadot-sdk/master/frame_support/traits/trait.ExecuteBlock.html#tymethod.execute_block Although it isn't documented, they also panic for inherents out of order which is the same thing I'm doing here. https://github.com/paritytech/polkadot-sdk/blob/7aace06b3d653529ab992c8ff96571d54bf00a36/substrate/frame/executive/src/lib.rs#L491-L493

node/src/service.rs
```diff
@@ -242,16 +243,27 @@ pub fn new_full(config: Configuration) -> Result<TaskManager, ServiceError> {
...
-							*timestamp,
-							slot_duration,
-						);
-
-                    Ok((slot, timestamp))
+                create_inherent_data_providers: move |parent_hash, ()| {
+                    let parent_block = client_for_cidp
+                        .clone()
+                        .block(parent_hash)
+                        .expect("a")
```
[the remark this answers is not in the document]
muraca: It took a while, but I found where this is used. 
https://github.com/paritytech/polkadot-sdk/blob/7aace06b3d653529ab992c8ff96571d54bf00a36/substrate/client/consensus/slots/src/slots.rs#L154 
It logs the error and tries again next slot, so it might not like a panic

JoshOrndorff:
I've discussed this in element with @muraca and he thought of a nice way to simplify the Timestamp piece. Rather than having separate noted and best timestamps, we will have a single notion of timestamp that also includes the block number in which it was stored.

```Rust
pub struct Timestamp {
  time: u64,
  block: u64,
}
```

Then timestamp update transactions, only need to peek at a the previous timestamp from the parent block (not consume anything). We can ensure that it is indeed the most recent timestamp from the previous block by checking it's stored block number.

JoshOrndorff:
We also discussed the wallet issue which CI is indeed catching. I believe it exists also on main, and it is just not immediately obvious because most blocks do not have transactions on `main` whereas in this branch every transaction has a timestamp inherent. This is likely related to the same upstream change that casused #119.

node/src/service.rs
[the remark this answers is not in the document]
JoshOrndorff: Definitely we will not be able to author a block if the database does not know about the previous block. What do you suggest to improve this?

tuxedo-core/src/constraint_checker.rs
```diff
@@ -34,32 +31,43 @@ pub trait SimpleConstraintChecker: Debug + Encode + Decode + Clone {
 /// A single constraint checker that a transaction can choose to call. Checks whether the input
 /// and output data from a transaction meets the codified constraints.
 ///
-/// This full ConstraintChecker should only be used if there is more that a piece wants to do such
-/// as check the verifier information in some unique way.
+/// This full ConstraintChecker should only be implemented if the piece logic cannot be expressed with
+/// the SimpleConstraintChecker. For example, if you need to enforce the verifier is a particular type
+/// or contains a certain value. Another reason would be if you need to implement an inherent.
```
muraca, line 36: I was thinking maybe we could add an `InherentConstraintChecker` trait instead.

wardrobe/timestamp/src/lib.rs
```diff
@@ -167,45 +159,32 @@ impl<T: TimestampConfig + 'static, V: Verifier + From<UpForGrabs>> ConstraintChe
...
-            .extract::<NotedTimestamp>()
-            .map_err(|_| Self::Error::BadlyTyped)?
-            .0;
-
-        // Make sure there are no extra outputs
+            .extract::<Timestamp>()
+            .map_err(|_| Self::Error::BadlyTyped)?;
         ensure!(
-            output_data.len() == 2,
+            output_data.len() >= 2,
```
muraca, line 175: I thought the output of the inherent was supposed to have length 1.
Maybe we can simplify the check, and instead of having two ensure!, we could only check if the length is 1.

wardrobe/timestamp/src/lib.rs
```diff
@@ -47,22 +44,18 @@ mod update_timestamp_tests;
 /// A piece-wide target for logging
 const LOG_TARGET: &str = "timestamp-piece";
 
-/// A timestamp, since the unix epoch, that is the latest time ever seen in the history
-/// of this chain.
+/// A timestamp, since the unix epocht, noted at some point in the history of the chain.
```
muraca, line 47: epocht 😄 
(for some reason I can't suggest changes)

muraca (commented):
New changes with single-output design are generally good, let's wait for the tests to see if I missed anything

tuxedo-core/src/types.rs
```diff
@@ -47,6 +47,20 @@ pub struct Transaction<V: TypeInfo, C: TypeInfo> {
     pub checker: C,
 }
 
+impl<V: Clone, C: Clone> Transaction<V, C> {
```
coax1d, line 50: Not sure if this is better abstracted over a trait but since you cant use `From` here this is not a bad solution

tuxedo-core/src/types.rs
```diff
@@ -94,18 +108,36 @@ impl<V: Decode + TypeInfo, C: Decode + TypeInfo> Decode for Transaction<V, C> {
...
     type Call = Self;
     type SignaturePayload = ();
 
     fn new(data: Self, _: Option<Self::SignaturePayload>) -> Option<Self> {
         Some(data)
     }
 
-    // This function has a default implementation that returns None.
-    // TODO what are the consequences of returning Some(false) vs None?
+    // The signature on this function is not the best. Ultimately it is
```
coax1d, line 123: Perhaps as a refactor not now necessarily we represent this logic using an Enum since the variants are then more clear and concise. Interpreting these three scenarios in its present form is not totally ideal but understand why you did it this way.

wardrobe/poe/src/lib.rs
coax1d: If we dont know why we need it and it doesnt stop us from doing what we are doing Id say best not to have it tbh.

wardrobe/timestamp/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+version = "0.1.0"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+log = { workspace = true }
+parity-scale-codec = { features = [ "derive" ], workspace = true }
+scale-info = { features = [ "derive" ], workspace = true }
+serde = { features = [ "derive" ], optional = true, workspace = true }
+sp-api = { default_features = false, workspace = true }
```
coax1d, line 14: Just as an FYI if in the Workspace we automatically add 'default-features = false' then we dont also need to do it here. We instead can just do 'sp-api = { workspace = true }'

wardrobe/timestamp/src/cleanup_tests.rs
```diff
@@ -0,0 +1,178 @@
...
+fn cleanup_timestamp_input_newer_than_reference() {
+    let old = NotedTimestamp(200);
+    let newer = NotedTimestamp(100);
+
+    let input_data = vec![old.into()];
+    let peek_data = vec![newer.into()];
+
+    assert_eq!(
+        CleanUpTimestamp::<AlwaysBlockTwo>::default().check(&input_data, &peek_data, &[]),
+        Err(DontBeSoHasty)
```
coax1d, line 58: Great name love it lol

wardrobe/timestamp/src/lib.rs
```diff
@@ -0,0 +1,382 @@
...
+    /// block times which is a common default in Substrate chains because of Polkadot.
+    const MIN_BLOCKS_BEFORE_CLEANUP: u32 = 15_000;
+}
+
+/// Reasons that setting or cleaning up the timestamp may go wrong.
+#[derive(Debug, Eq, PartialEq)]
+pub enum TimestampError {
+    /// UTXO data has an unexpected type
+    BadlyTyped,
+
```
coax1d, line 98: Extra space

wardrobe/timestamp/src/lib.rs
```diff
@@ -0,0 +1,382 @@
...
+    TooManyOutputsWhileSettingTimestamp,
+    /// The previous timestamp that is peeked at must be from the immediate ancestor block, but this one is not.
+    PreviousTimestampWrongHeight,
+    /// No previous timestamp was peeked at in this transaction, but at least one peek is required.
+    MissingPreviousTimestamp,
+    /// Inputs were specified while setting the timestamp, but none are allowed.
+    InputsWhileSettingTimestamp,
+    /// The new timestamp is not sufficiently far after the previous (or may even be before it).
+    TimestampTooOld,
+
```
coax1d, line 113: Extra space

wardrobe/timestamp/src/lib.rs
```diff
@@ -0,0 +1,382 @@
...
+        input_data: &[tuxedo_core::types::Output<V>],
+        peek_data: &[tuxedo_core::types::Output<V>],
+        output_data: &[tuxedo_core::types::Output<V>],
+    ) -> Result<TransactionPriority, Self::Error> {
+        log::debug!(
+            target: LOG_TARGET,
+            "🕰️🖴 Checking constraints for SetTimestamp."
+        );
+
+        // FRAME pallet authors are required to ensure_none! to make sure this is an inherent.
```
coax1d, line 158: This comment might not make so much sense if the person using Tuxedo has never used FRAME but I dont mind it all that much

wardrobe/timestamp/src/lib.rs
```diff
@@ -0,0 +1,382 @@
...
+    for SetTimestamp<T>
+{
+    type Error = sp_timestamp::InherentError;
+    const INHERENT_IDENTIFIER: sp_inherents::InherentIdentifier = sp_timestamp::INHERENT_IDENTIFIER;
+
+    fn create_inherent(
+        authoring_inherent_data: &InherentData,
+        previous_inherent: Option<(Transaction<V, Self>, H256)>,
+    ) -> tuxedo_core::types::Transaction<V, Self> {
+        // Extract the current timestamp from the inherent data
```
coax1d, line 236: This variable could be renamed to curr_timestamp and then this comment would be unnecessary

wardrobe/timestamp/src/lib.rs
```diff
@@ -0,0 +1,382 @@
...
+
+                // We are given the entire previous inherent in case we need data from it or need to scrape the outputs.
+                // But out transactions are simple enough that we know we just need the one and only output.
+                peeks.push(OutputRef {
+                    tx_hash: previous_id,
+                    // There is always 1 output, so we know right where to find it.
+                    index: 0,
+                });
+            }
+        }
```
coax1d, line 269: This whole block is the consequence of this tuple matching decision. Not my favorite.

wardrobe/timestamp/src/lib.rs
```diff
@@ -0,0 +1,382 @@
...
+        let on_chain_timestamp = inherent.outputs[0].payload.extract::<Timestamp>().expect(
+            "SetTimestamp extrinsic should have an output that decodes as a StorableTimestamp.",
+        );
+
+        log::debug!(
+            target: LOG_TARGET,
+            "🕰️🖴 In-block timestamp is: {:#?}", on_chain_timestamp
+        );
+
+        // Although FRAME makes the check for the minimum interval here, we don't.
```
coax1d, line 309: Again references to FRAME might not be necessary since users of Tuxedo can come from other places not caring about FRAME

wardrobe/timestamp/src/lib.rs
```diff
@@ -0,0 +1,382 @@
...
+            checker: Self::default(),
+        }
+    }
+
+    fn check_inherent(
+        importing_inherent_data: &InherentData,
+        inherent: Transaction<V, Self>,
+        result: &mut CheckInherentsResult,
+    ) {
+        // Extract the local view of time from the inherent data
```
coax1d, line 289: This comment is obvious I would remove it

coax1d (changes requested):
This was a big one but nice job figuring out how to do this!

wardrobe/timestamp/src/lib.rs
muraca: fixed part 1 in 10213120bb46a1a3e6cc48d25907d14540f292b4

wardrobe/timestamp/src/lib.rs
muraca: It will be completely removed once we can put the extrinsic in genesis block.
@JoshOrndorff can you remind me why it would be a problem if we put the timestamp in the genesis block as raw data rather than an extrinsic?

tuxedo-core/src/types.rs
muraca: Keep in mind that this is the implementation of a Substrate trait.
See #109 for more info

wardrobe/timestamp/src/lib.rs
muraca: fixed in 6409334ff59118cda4dc786271e5e6ece396bd23

wardrobe/timestamp/src/lib.rs
muraca: fixed in 6409334ff59118cda4dc786271e5e6ece396bd23

wardrobe/timestamp/src/lib.rs
muraca: fixed in 6409334ff59118cda4dc786271e5e6ece396bd23

tuxedo-core/src/executive.rs
```diff
@@ -360,14 +389,85 @@ impl<
...
+
+    pub fn check_inherents(block: B, data: InherentData) -> sp_inherents::CheckInherentsResult {
+        debug!(
+            target: LOG_TARGET,
+            "Entering `check_inherents`"
+        );
+
+        let mut result = CheckInherentsResult::new();
+
+        // Tuxedo requires that all inehrents come at the beginning of the block.
```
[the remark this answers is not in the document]
muraca: 6409334ff59118cda4dc786271e5e6ece396bd23

tuxedo-core/src/executive.rs
```diff
@@ -360,14 +389,85 @@ impl<
...
+            "Entering `inherent_extrinsics`."
+        );
+
+        // Extract the complete parent block from the inheret data
+        let parent: B = data
+            .get_data(&PARENT_INHERENT_IDENTIFIER)
+            .expect("1")
+            .expect("2");
+
+        // Extract the beginning-of-block inherents from the previous block.
```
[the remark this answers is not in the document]
muraca: improved in 6409334ff59118cda4dc786271e5e6ece396bd23

wardrobe/timestamp/src/cleanup_tests.rs
```diff
@@ -1,102 +1,120 @@
 //! Unit tests for the Timestamp piece.
 //! This module tests the secondary flow of cleaning up old timestamps.
 
-use super::{
-    BestTimestamp, CleanUpTimestamp, NotedTimestamp, SimpleConstraintChecker, TimestampConfig,
-    TimestampError,
-};
+use super::*;
```
JoshOrndorff, line 4: I used the longer imports so that I _didn't_ get the `ConstraintChecker` trait in scope. By only having `SimpleConstraintChecker` in scope, you don't need to specify which trait you're trying to call in each test.

tuxedo-core/aggregator/src/lib.rs
```diff
@@ -115,12 +130,28 @@ pub fn tuxedo_constraint_checker(attrs: TokenStream, body: TokenStream) -> Token
...
     let error_type = Ident::new(&error_type_name, outer_type.span());
-    let inner_types = inner_types.clone();
+
+    let mut inherent_hooks_name = outer_type.to_string();
+    inherent_hooks_name.push_str("InherentHooks");
+    let inherent_hooks = Ident::new(&inherent_hooks_name, outer_type.span());
+
+    let vis = ast.vis;
+
+    // TODO there must be a better way to do this, right?
```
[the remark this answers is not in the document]
JoshOrndorff: I'm fine keeping this ugly todo for now. But borrowing is a good idea. I honestly didn't think of that. Let me try it.

tuxedo-core/src/executive.rs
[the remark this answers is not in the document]
JoshOrndorff: > So this is extracting the end of block inherents from the previous block as the beginning of block inherents for this block?

No, that's not what's happening. end-of-block inherents are not implemented yet and aren't required for parachain support. Anything related to end-of-block inherents will come in the future.

This logic is extracting all of the inherents from the previous block. But they will not go into the next block. Once the previous inherents are extracted, they are passed into the `create_extrinsics` runtime api, an off-chain runtime api that is used for creating this blocks inherents.

So why do we need the old inherents to create the new ones? Because we need to know what outputs it created so the next block's inherent can use them as inputs or peeks.

tuxedo-core/src/types.rs
JoshOrndorff: Yeah, I agree this should be an enum. Now that this PR is mostly ironed out and with the research from #109 I think we should make a PR to Substrate.

wardrobe/timestamp/Cargo.toml
JoshOrndorff: Nice! I made an issue for it https://github.com/Off-Narrative-Labs/Tuxedo/issues/122

wardrobe/poe/src/lib.rs
JoshOrndorff: Agreed. Removed in 8c09593bd7cbf03e3228dc00f4ff30379ddd6979

tuxedo-core/aggregator/src/lib.rs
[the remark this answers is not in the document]
JoshOrndorff: `cannot move out of `*variants` which is behind a shared reference`

Since none of us are macro experts and it is working right now, I'm going to leave it. Hopefully someday we'll get a real macro expert to take a look.

wardrobe/timestamp/src/lib.rs
JoshOrndorff: 730028a0ee7398213d3b79abdca209ff5356cb34

wardrobe/timestamp/src/cleanup_tests.rs
muraca: It's better to do how it was done before, sorry for the mess. Should I fix it, or are you taking care of this?

wardrobe/timestamp/src/lib.rs
JoshOrndorff: First regarding the code style and readability. I agree it isn't the easiest thing in the world to read. It would actually look a lot more idiomatic if I removed all the comments, but I thought they helped explain the flow. An alternative is to use `if` `else if` `else` but we definitely need these three branches at least until we have a genesis timestamp.

Now regarding the genesis timestamp. My main motivation for putting a timestamp extrinsic in the genesis block is so that we don't need a the first-block-hack in the inherent logic. It isn't because I really want the genesis block to be timestamped. So the genesis timestamp needs to be where the inherent logic expects it, as an extrinsic.

And as a followup point, where else would you put it? The only other place to jam arbitrary data is in the header's digest field.

wardrobe/timestamp/src/lib.rs
JoshOrndorff: Yeah, I hear your point, and I took it on the other comment. But I do think it is a good idea to document the places where we take different design decisions than FRAME. I think it is more appropriate in a piece like this that directly corresponds to a specific frame pallet. I'm down to remove it though.

tuxedo-core/src/executive.rs
```diff
@@ -360,14 +389,85 @@ impl<
...
+    pub fn check_inherents(block: B, data: InherentData) -> sp_inherents::CheckInherentsResult {
+        debug!(
+            target: LOG_TARGET,
+            "Entering `check_inherents`"
+        );
+
+        let mut result = CheckInherentsResult::new();
+
+        // Tuxedo requires that all inehrents come at the beginning of the block.
+        // (Soon we will also allow them at the end, but never throughout the body.)
```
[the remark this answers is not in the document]
JoshOrndorff: 8e7350e4f2929207c7559c3913867615114e2e94

wardrobe/timestamp/src/cleanup_tests.rs
JoshOrndorff: I'm not working on it. I also don't really care that much. Just left the comment for your understanding. Feel free to change it back if you want, but not a big deal for me.

muraca: I think it should be considered a good practice, since it keeps the code both more coincise and more readable, and it should be kept as an example for posterity

muraca: solved in 52b2f855ba30ce70a1ea2ca70ad6583b29bcc19e

tuxedo-core/src/executive.rs
```diff
@@ -291,8 +303,25 @@ impl<
...
         sp_io::storage::set(HEADER_KEY, &block.header().encode());
 
+        // Tuxedo requires that inherents are at the beginning (and soon end) of the
+        // block and not scattered throughout. We use this flag to enforce that.
+        let mut finished_with_opening_inherents = false;
+
         // Apply each extrinsic
         for extrinsic in block.extrinsics() {
+            // Enforce that inherents are in the right place
+            match (
```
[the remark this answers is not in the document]
muraca: you might compress `!a && b` and `a && !b` using a XOR, since both result into  `=> (), `

or you might go hard and do (untested)
```rust
if a == b {
    if a {
        panic!("Tried to execute opening inherent after switching to non-inherents.");
    }
    finished_with_opening_inherents = true, 
}
```

I think the match is more readable 😄

wardrobe/timestamp/src/lib.rs
muraca: > where else would you put it? 

I might not be right, but we initialize storage with UTXOs that contain money, could we do the same with a timestamp? 
However I think the value would be hard-coded so I'm not sure it would work well in real use cases - you should know before compiling the timestamp of when you will run the chain for the first time.

tuxedo-core/src/executive.rs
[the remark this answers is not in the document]
JoshOrndorff: Okay, you guys inspired me and I revised this in 87ee4dca4a517b418457195e86053a59ca7b5195

wardrobe/timestamp/src/lib.rs
JoshOrndorff: Oh, I see what you mean. Like put the timestamp UTXO in the genesis storage but no actual timestamp extrinsic in the genesis block.

I don't think that would fully solve the problem. It's true that there would be a timestamp there to peek at in block 1. That is good.

But now consider you are the node that needs to author block 1. How do you know what OutputRef to peek at in your timestamp inherent? The usual way is to scrape the previous block to find it. But if there is no transaction in the previous (genesis) block there is nothing to scrape.

tuxedo-core/src/executive.rs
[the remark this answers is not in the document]
muraca: I'm not surprised that there was a simpler solution. I'm tired lol

wardrobe/timestamp/src/lib.rs
JoshOrndorff: Yeah, good call. I had this wrong.

tuxedo-core/src/constraint_checker.rs
JoshOrndorff: Yeah. This is a good idea. It would simplify a lot of things. This is roughly what I had in mind when I put  `(Deferred to followup) Rework relationship between constraint checkers and inherent hooks.` as one of the checklist items in the PR description.

node/src/service.rs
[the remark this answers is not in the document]
muraca: I was trying earlier to make this return an error when needed, instead of panicking. I wanted to do someting like:
```rust
let parent_block = client_for_cidp
    .clone()
    .block(parent_hash)?
    .ok_or(sp_blockchain::Error::UnknownBlock(parent_hash))?
    .block;
```
There was some issues with the closure being async, and I'm not familiar with that, so please try to make this work without spending too much time on it, otherwise it's fine to leave it like this.

tuxedo-core/aggregator/src/lib.rs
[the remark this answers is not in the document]
gabriele-0201: The first problem here is that the value can't be cloned in the quote macro (as far as I know, if it would be possible with some magic trick then the solution would be obvious) so all the clones need to be done before the macro, and that repeated code is the result. An easy solution is to involve a macro_rules to avoid repeat all the syntax but the inner problem is that inside macro_rules identifiers can't be created, then I came up with two solutions:
1. the first one need to provide to the macro the name of the variables and the name of the variable that needs to be copied, but this result into more lines and even uglier then the repeated code
2. Using the `paste` macro, that enables you to append identifiers to create new ones, the code is the following:

```rust
    macro multiple_assign {
        ($($num:literal),* = $var:ident) => {
            paste!{ $(let [<$var $num>] = $var.clone();)* }
        }
    }
    
    multiple_assign!(2,3,4,6 = inner_types);
    multiple_assign!(2,3,4,5,6 = variants);
```

the final code is not smaller than the previous one but is more scalable if the required copies of inner_types and variants will grow in the future

PR #104, opened 2023-09-29
Add docs to PoE config trait

JoshOrndorff:
Simple PR adds some missing doc comments


wardrobe/poe/src/lib.rs
```diff
@@ -70,6 +70,8 @@ pub enum ConstraintCheckerError {
     EffectiveHeightInPast,
 }
 
+/// Configuration items for the Prrof of Existence piece when it is
```
JoshOrndorff, line 73:
```suggestion
/// Configuration items for the Proof of Existence piece when it is
```

PR #105, opened 2023-09-29
Accumulators for intra-block book-keeping

JoshOrndorff:
## Overview

This PR adds a feature that is called "accumulators" in the code and I've also considered calling "scratchpads". An accumulator serves as a temporary place for Tuxedo pieces to do some accounting within a single block. The accumulator will be cleared automatically by the Tuxedo executive at the end of each block, so the stored data will not be available in the next block, or even after the block's execution closes.

## Usecases

1. Tallying up the tips from each transaction so that the block author can claim them at the end of each block.
2. Making sure that inherents are set exactly once in each block.
3. No other concrete ones yet, but I suspect there will be.

## Fit in Substrate

This feature is similar to https://github.com/paritytech/polkadot-sdk/issues/359 in some ways. However it is implemented in the Tuxedo core which is entirely in the runtime. That issue asks for host functions and performance which this will not provide. This also does not satisfy @burdges's request to have the data in-memory and skip scale encoding.

## Drawbacks

This feature is pretty un-UTXO-y in the sense that we are basically providing (a strict structured interface around) storage items to the Tuxedo pieces. I've tried very hard to avoid storage items. I even went to great lengths to handle inherents in a UTXO-native way in #100 and I'm glad I did.

The thing that makes me mostly okay with this is that the storage is cleared at the end of each block, so the data does not persist. I think the term "scratchpad" emphasizes this well as it is a temporary place to write some notes while doing a job and it is not part of the final project.

I have not yet thought of another way to achieve giving tips to block authors, a feature that has existed even since bitcoin. Indeed, even in bitcoin there is a dedicated variable for tallying such tips.

## Refactor and Features

@clangenb I remember you saying you like it when the refactor goes in first and the feature that motivated comes in another PR. That has stuck with me. Once I get this ironed out, I plan to separate the core feature from the way I'm using it in the money piece, and put them in separately. Thanks for inspiring more discipline.


burdges:
As an aside, I'd think "pallets" could work like this:
```
use spin::Mutex; // https://docs.rs/spin/latest/spin/index.html

struct PalletAcc { ... }

static MUTEX: Mutex<Option<PalletAcc>> = Mutex::new(None);

fn on_initialize() {
    let lock = MUTEX.try_lock().expect("Reentrancy bug!");
    *lock = Some(PalletAcc::new())
}

fn verify_tx() {
    // Setup validation
    let lock = MUTEX.try_lock().expect("Reentrancy bug!");
    match *lock {
        None => {
            // we're not inside execute_block so do full tx validation, abort if fails
        },
        Some(acc) => { 
            // we're inside execute_block so use acc: &mut PalletAcc
        }.
    }
    drop(lock);
    // Finish validation, abort if fails
}

fn on_finalize() {
    let lock = MUTEX.try_lock().expect("Reentrancy bug!");
    let acc = lock.take()
    drop(lock);
    // Validate acc itself, abort if fails
}
```
We do not require the spin lock per se, since the runtime is single threaded, but whatever.

tuxedo-core/src/constraint_checker.rs
```diff
@@ -99,14 +188,19 @@ pub mod testing {
...
             _peek_data: &[DynamicallyTypedData],
             _output_data: &[DynamicallyTypedData],
-        ) -> Result<TransactionPriority, ()> {
+        ) -> Result<ConstraintCheckingSuccess<()>, ()> {
             if self.checks {
-                Ok(0)
+                Ok(ConstraintCheckingSuccess {
+                    priority: 0,
+                    accumulator_value: (),
+                })
```
JoshOrndorff, line 203: I guess repeating this is kind of annoying. Maybe we should have an impl like

```rust
impl From<TransactionPrtiority> for ConstraintCheckingSuccess<()> {

  fn from(priority: TransactionPriority) -> Self {
    Self {
      priority,
      accumulator_value: (),
    }
  }
}
```

JoshOrndorff:
Oh crap, I haven't actually updated to executive to use the new accumulator logic. So there is still that part. But anyway this is coming together now, and some review on the design is very welcome.

tuxedo-core/src/constraint_checker.rs
clangenb: Haha, I am flattered by the credits. :)

I can definitely recommend doing these boilerplate implementations - 0 cost abstractions; hence, 0 reason not to do so, haha.

muraca:
@JoshOrndorff I hope this commit is helpful. 
The From/Into implementation for ConstraintCheckingSuccess was not feasible, so I added a new method that does exactly that, please use that one instead of the ugly conversion in the proc macro

tuxedo-core/src/constraint_checker.rs
```diff
@@ -7,11 +7,111 @@ use sp_std::{fmt::Debug, vec::Vec};
...
+/// Some typical usecases for an accumulator would be making sure that the block author reward
+/// does not exceed the block's transaction fees. Or making sure that a timestamp transaction occurs
+/// at most once in a single block.
+///
+/// The accumulator is reset automatically at the end of every block. It is not possible to store
+/// data here for use in subsequent blocks.
+pub trait Accumulator {
+    /// The type that is given and also the type of the accumulation result.
+    /// I realize that the most general accumulator swill use two different types for those,
+    /// but let's do that iff we ever need it. I probably will need to so I can do a simple counter.
```
coax1d, line 72:
```suggestion
    /// but let's do that iff we ever need it. Probably will need to so a simple counter can be implemented.
```

I would avoid comments that reference yourself as I.

tuxedo-core/src/constraint_checker.rs
```diff
@@ -7,11 +7,111 @@ use sp_std::{fmt::Debug, vec::Vec};
...
+    fn initial_value(_: Self::ValueType) -> Self::ValueType;
+
+    // TODO This needs to be improved because keeping them unique is kind of a footgun right now.
+    // It is a function and not a const because in aggregate runtimes, we will need to determine,
+    // at the executive level, what the key is, which means we need to match in the aggregate implementation.
+    /// A unique key for this accumulator in the runtime. Like with storage types,
+    /// Runtime authors must take care that this key is not used anywhere else in the runtime.
+    ///
+    /// This is a function and takes a value, as opposed to being a constant for an important reason.
+    /// Aggregate runtimes made from multiple pieces will need to give a different initial value depending
```
coax1d, line 90:
```suggestion
    /// Aggregate runtimes made from multiple pieces will need to give a different key path depending
```

tuxedo-core/src/constraint_checker.rs
```diff
@@ -7,11 +7,111 @@ use sp_std::{fmt::Debug, vec::Vec};
...
+    /// which of the constituent constraint checkers is being called.
+    fn key_path(_: Self::ValueType) -> &'static [u8];
+
+    /// This function is responsible for combining or "folding" the intermediate value
+    /// from the current transaction into the accumulatoed value so far in this block.
```
coax1d, line 95:
```suggestion
    /// from the current transaction into the accumulated value so far in this block.
```

tuxedo-core/src/constraint_checker.rs
```diff
@@ -22,13 +122,16 @@ pub trait SimpleConstraintChecker: Debug + Encode + Decode + Clone {
...
+    type Accumulator: Accumulator;
+
     /// The actual check validation logic
     fn check(
         &self,
         input_data: &[DynamicallyTypedData],
         peek_data: &[DynamicallyTypedData],
         output_data: &[DynamicallyTypedData],
-    ) -> Result<TransactionPriority, Self::Error>;
+    ) -> Result<ConstraintCheckingSuccess<<Self::Accumulator as Accumulator>::ValueType>, Self::Error>;
```
coax1d, line 134: I think type aliasing for stuff like this is preferred. Things can start to look really ugly without it..

tuxedo-core/src/executive.rs
```diff
@@ -164,18 +165,39 @@ impl<
...
-            valid_transaction.requires.is_empty(),
+            prelim_valid_transaction.requires.is_empty(),
             UtxoError::MissingInput
         );
 
+        // Process the accumulator
+        let acc_key = C::Accumulator::key_path(
+            <C::Accumulator as Accumulator>::ValueType::decode::<&[u8]>(&mut &Vec::<u8>::new()[..])
+                .expect("TODO!"),
+        ); // TODO what should this key be?
```
coax1d, line 182: Do we have the answer to this question?

tuxedo-core/src/types.rs
```diff
@@ -132,6 +136,9 @@ pub enum UtxoError<ConstraintCheckerError> {
     VerifierError,
     /// One or more of the inputs required by this transaction is not present in the UTXO set
     MissingInput,
+    // TODO set a nicer description for this.
```
coax1d, line 139: Can complete this TODO or it still is lingering?

tuxedo-core/src/types.rs
```diff
@@ -148,6 +155,46 @@ pub struct Output<V> {
     pub verifier: V,
 }
 
+/// Similar to Substrate's ValidTransaction. This type indicates that a transaction
+/// has passed the standard UTXO checks and is either valid now or may be in the future.
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Encode, Decode, Debug, PartialEq, Eq, Clone, TypeInfo)]
+pub struct PreliminarilyValidTransaction<A> {
+    pub provides: Vec<Vec<u8>>,
+    pub requires: Vec<Vec<u8>>,
```
coax1d, line 164: any better description for what these` Vec<Vec<u8>>` are? An alias of any kind? if not is what it is.

wardrobe/money/src/lib.rs
```diff
@@ -84,87 +68,148 @@ pub enum ConstraintCheckerError {
...
+                .ok_or(ConstraintCheckerError::ValueOverflow)?;
+        }
+
+        ensure!(
+            total_output_value <= total_input_value,
+            ConstraintCheckerError::OutputsExceedInputs
+        );
+
+        // Priority is based on how many token are burned
+        // Type stuff is kinda ugly. Maybe division would be better?
```
coax1d, line 154: Still a relevant comment?

wardrobe/money/src/lib.rs
```diff
@@ -84,87 +68,148 @@ pub enum ConstraintCheckerError {
...
+        let burned = total_input_value - total_output_value;
+
+        // Just a saturated version of the burned amount.
+        // TODO can't I use saturating_into or something here instead of writing it myself?
```
coax1d, line 158: Can Todo or personal comment be addressed or removed?

tuxedo-core/src/constraint_checker.rs
```diff
@@ -7,11 +7,111 @@ use sp_std::{fmt::Debug, vec::Vec};
...
+    /// but let's do that iff we ever need it. I probably will need to so I can do a simple counter.
+    type ValueType: Debug + Encode + Decode;
+
+    /// The accumulator value that should be used to start a fresh accumulation
+    /// at the beginning of each new block.
+    ///
+    /// This is a function and takes a value, as opposed to being a constant for an important reason.
+    /// Aggregate runtimes made from multiple pieces will need to give a different initial value depending
+    /// which of the constituent constraint checkers is being called.
+    fn initial_value(_: Self::ValueType) -> Self::ValueType;
```
coax1d, line 81: Will this just always be the `Default` for the `ValueType` associated type? Should `ValueType` just instead implement the `Default` trait?

coax1d (changes requested):
Looks good but lots of comment clean ups with regards to Todos and some other questions I had

tuxedo-core/src/executive.rs
JoshOrndorff: The `acc_key` should be something that one of the pieces' accumulators returns. For example `imbalance` from the money piece or `stub_acc` from the `()` implementation.

The prefixed one right below would then be `:accumulator:imbalance`. The point of prefixing them is that we can kill them all in one fell swoop at the end of the block.

tuxedo-core/src/types.rs
JoshOrndorff:
```suggestion
```

Yeah, the existing one seems good to me.

tuxedo-core/src/types.rs
```diff
@@ -155,10 +160,39 @@ pub struct Output<V> {
...
+impl<A> From<ValidTransaction> for PreliminarilyValidTransaction<A> {
+    fn from(transaction: ValidTransaction) -> Self {
+        Self {
+            provides: transaction.provides,
+            requires: transaction.requires,
+            priority: transaction.priority,
+            intermediate_accumulator_value: None,
+        }
+    }
+}
```
JoshOrndorff, line 190: @muraca What is the point of this one? When would we need to go in this reverse direction from valid to preliminarily valid?

wardrobe/money/src/lib.rs
```diff
@@ -95,8 +95,8 @@ impl Accumulator for ImbalancedFundsAccumulator {
         0
     }
 
-    fn key_path(_: Self::ValueType) -> &'static str {
-        "imbalanc"
+    fn key_path(_: Self::ValueType) -> &'static [u8] {
+        b"imbalanc"
```
JoshOrndorff, line 99: imbalance with an e

Previously I was using `[u8; 8]` and that's why the e was missing, but we can now use the full word.

tuxedo-core/aggregator/src/lib.rs
```diff
@@ -115,20 +115,39 @@ pub fn tuxedo_constraint_checker(attrs: TokenStream, body: TokenStream) -> Token
...
     let inner_types = inner_types.clone();
+    let inner_types2 = inner_types.clone();
+    let inner_types3 = inner_types.clone();
+    let inner_types4 = inner_types.clone();
+    let inner_types5 = inner_types.clone();
     let variants2 = variants.clone();
+    let variants3 = variants.clone();
+    let variants4 = variants.clone();
+    let variants5 = variants.clone();
+    let variants6 = variants.clone();
```
JoshOrndorff, line 141: There's gotta be a better way to do this right? Does anyone know? I'm pretty much a macro newb. This crate is by far the most complex one I've ever written.

tuxedo-core/aggregator/src/lib.rs
```diff
@@ -138,18 +157,80 @@ pub fn tuxedo_constraint_checker(attrs: TokenStream, body: TokenStream) -> Token
...
                 peeks: &[tuxedo_core::types::Output<#verifier>],
                 outputs: &[tuxedo_core::types::Output<#verifier>],
-            ) -> Result<TransactionPriority, Self::Error> {
+            ) -> Result<tuxedo_core::constraint_checker::ConstraintCheckingSuccess<<Self::Accumulator as tuxedo_core::constraint_checker::Accumulator>::ValueType>, Self::Error> {
                 match self {
                     #(
-                        Self::#variants2(inner) => inner.check(inputs, peeks, outputs).map_err(|e| Self::Error::#variants2(e)),
+                        Self::#variants6(inner) => inner.check(inputs, peeks, outputs)
+                        .map(|old| {
+                            //TODO I would really rather have an into or from impl for ConstraintCheckingSuccess, but that's just won't compile
```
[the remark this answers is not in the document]
JoshOrndorff: Yes, I was stuck/confused on this last night. Matteo's `transform` method is a good way forward.

tuxedo-core/src/constraint_checker.rs
JoshOrndorff: I thought about that same thing. All the examples of accumulators that I thought of can indeed be implemented using the default value as the initial value.

I ended up going with a separate method for these reasons:
1. Semantics. While it seems the initial value of the accumulator will typically be the default, there is no fundamental reason it must be this way. Someone may write an intuitive impl of `Default` for that purpose.
2. To mirror https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fold
3. The main one: For aggregate runtimes. When the constraint checker is composed with the macro from individual constraint checkers, it needs to do an actual match and return a different initial value depending on which piece is being called. Default does allow taking a param and returning a different result based on that param.

wardrobe/money/src/lib.rs
JoshOrndorff: It is still relevant to the code, but it is not introduced in this PR. This code block was already here and git just didn't notice it's the same. I support addressing this, but not in this PR.

wardrobe/money/src/lib.rs
JoshOrndorff:
```suggestion
```

I guess going with the manual one is fine actually. Unless someone can easily make `ssaturated_into` work, let's just leave this.

wardrobe/money/src/lib.rs
JoshOrndorff: Addressed in fa797b50f1d8fe70581ea2e29203472c08ca4758

wardrobe/money/src/lib.rs
JoshOrndorff: addressed in 3de2e96d5718e1714b17eaae8b87ce8efe0f6da6

JoshOrndorff:
Thanks for the reviews everyone. I know this PR is pretty far along, but I thought of an alternative design for tallying up fees and giving them to the block author and I need to write it out so we can consider them. My biggest complaint about the accumulators was always that they were not very UTXO native and were more like storage items, and this alternative design solves that problem.

When users send transactions, they must ensure that the input coins exceed the value of the output coins. They also have the ability to protect those output coins with whatever verifiers they want. Typically it might require a signature from the intended recipient. Of course token senders can also specify outputs that are UpForGrabs and can be claimed by anyone. And this is the key to the alternate design.

When you send tokens, you explicitly specify a tip to the block author by putting UpForGrabs. Only tokens that are explicitly included in an output as UpForGrabs may be claimed by the block author. If the input value exceeds the output value, then the difference truly is burned.

Then block authors claim their reward by inserting a transaction at the end of the block that sweeps up all of those UpForGrabs coins that were created during the block. There is no risk that someone else will get to those coins first because the block author can simply drop any transactions that try to claim the UpForGrabs ones.

The implementation would roughly be that when authoring, the block author notes all the UpForGrabs coins in a temporary storage that is cleared before the block ends just like how we do the extrinsics now.

The key benefits are:
1. All the tip-related state is still in UTXOs, not in the accumulator.
2. Importing nodes don't need to know anything about the implementation of this feature. They only have to check the normal UTXO rules.

tuxedo-core/src/types.rs
muraca: I'll have to double check tomorrow, but I think I did this because it wasn't possible to use some of the ValidTransactionBuilder's fields

JoshOrndorff:
Okay, I'm going to close this one for now.

Block rewards can be done more elegantly by the alternate design described above, and enforcing the right number of inherent calls can be done in the offchain pre-checks.

If we ever decide we need it, we have the design done and can re-open this.

PR #110, opened 2023-10-03
Ci improvements

### LINKED ISSUE #106: Code Coverage in CI
Similar to https://github.com/paritytech/zombienet-sdk/pull/88

For some context on my experience with code coverage, see #40

muraca:
• renamed build.yml to ci.yml
• use toml-sort action, removed toml-sort.yml
• removed check
• test with nextest, llvm cov for coverage ~ fixes #106 
• added --no-deps to clippy
• bumped version for updated actions
• removed unmantained actions
• ⚠️  @JoshOrndorff ⚠️ tests and clippy start only after rustfmt and toml sort are successful, this **is intentional** and I would like to keep it


rust-toolchain.toml
```diff
@@ -0,0 +1,4 @@
+[toolchain]
+channel = "nightly"
+profile = "minimal"
+targets = [ "wasm32-unknown-unknown" ]
```
JoshOrndorff, line 4: Compare this to [Moonbeam's](https://github.com/moonbeam-foundation/moonbeam/blob/master/rust-toolchain)

```toml
[toolchain]
channel = "1.71.0"
components = [ "rustfmt", "clippy" ]
targets = [ "wasm32-unknown-unknown" ]
profile = "minimal"
```

A few questions:
1. Do we actually need nightly as the channel? If not, I prefer 1.72.0. I don't really know much about this though.
2. Should we list components like Moonbeams?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,80 @@
...
+    needs: [fmt, toml-sort]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install tooling
+        run: |
+          sudo apt-get install -y protobuf-compiler
+          protoc --version
+      - name: Install Clippy
+        run: rustup component add clippy
```
JoshOrndorff, line 76: If we have the components listed in the `rust-toolchain.toml` file can we remove this?

JoshOrndorff (approved):
Overall looks good. I left a few questions but I trust your judgement.

rust-toolchain.toml
muraca: I need to study the toolchain file and set the right parameters, I was hoping you were an expert in that matter. 
The current one reflects the settings I found in the old CI script

muraca:
@JoshOrndorff please set Workflow permissions to read and write in [Settings > Actions > General](https://github.com/Off-Narrative-Labs/Tuxedo/settings/actions).

JoshOrndorff:
For some reason, that option is disabled for me.

image

According to https://docs.github.com/en/actions/security-guides/automatic-token-authentication#example-1-passing-the-github_token-as-an-input it looks like how you did `pull_requests: write` should work.

muraca:
@JoshOrndorff something is still off

.github/workflows/ci.yml
muraca: It's working without it, not sure if it's because the components but I'll leave them there 😅

muraca:
The problem is that the action is performed on my fork, and the bot is trying to comment on this repo using the secret key of my personal repo. This does not happen if the branch is part of this repo, see #112 or https://github.com/muraca/Tuxedo/pull/2 . Looks like there's nothing we can do to avoid it, so I'll try to run the step only if the repo of the head and the base are the same, otherwise print it in the action output

muraca:
@JoshOrndorff please take a look at this https://github.com/Off-Narrative-Labs/Tuxedo/actions/runs/6411679096/job/17407615977?pr=110

JoshOrndorff:
So this is the one you want to merge, right?

Go ahead and merge the one you when you're ready, and close the other one.

Is it possible to print the results to terminal in both cases? (And also comment when possible?) If not no big deal.

muraca:
> Is it possible to print the results to terminal in both cases? (And also comment when possible?)  

yes, I am going to test it out right now.

muraca:
#112

PR #113, opened 2023-10-06
End-to-end CI for wallet

### LINKED ISSUE #101: End-to-end CI for wallet
Tuxedo strives for excellent test coverage. And we do have pretty good coverage in the core. However the wallet is mostly tested manually.

This issue is narrowly scoped for a concrete starting point. After this is achieved, the test suite can be expanded as much as we need.

This issue is complete when we have a CI job that starts up a Tuxedo template node, starts up a wallet, sends a token transfer transaction to the node, and confirms that the final state is as expected.

I want to avoid a complex CI setup because nobody who works on this project regularly is a CI expert. For example, I'd like to avoid introducing new languages like Typescript (even though it works very very well for Moonbeam, we are not TS experts and we want simplicity. Ideally this can be done with nothing more than github actions and popular Rust crates. If some bash scripting is necessary, that may be acceptable, but is not preferable.

muraca:
resolves #101

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,7 +90,7 @@ jobs:
 
   wallet:
     name: Wallet end-to-end test
-    # needs: [test, clippy]
+    needs: [test, clippy]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,7 @@ jobs:
         - name: Rust Cache
           uses: Swatinem/rust-cache@v2
           with:
+            cache-targets: true
             cache-on-failure: true
         - name: Run tests and print coverage data
           run: cargo llvm-cov nextest --json --output-path lcov.json
@@ -82,6 +83,7 @@ jobs:
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2
         with:
+          cache-targets: true
           cache-on-failure: true
       - name: Run Clippy
         run: cargo clippy --no-deps -- -D warnings
@@ -99,6 +101,7 @@ jobs:
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2
         with:
+          cache-targets: true
           cache-on-failure: true
       - name: Build
         run: cargo build --release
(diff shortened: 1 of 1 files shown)


muraca:
@JoshOrndorff as you can see, it takes too much time to build node and wallet, and cache does not seem to affect the building time that much, as it compiles basically everything anyway. 
At the moment, it depends on the correct execution of tests and clippy, I could make it run in parallel to those, but it would still be taking quite some time. 
We could consider running it only when something is pushed on master.

JoshOrndorff:
Great work. I know it's long but we need it. I honestly can't tell you how many times I've pushed a change knowing that the `tuxedo-core` was sound but wondering if I broke the wallet or block import in some subtle way. So however long it takes, we need it and we can work on getting the time down. And we need it on PRs, not just on `main` it is too late at that point. Also notice that the marginal cost to add more e2e tests once this goes in is very small. We only have to build the node once per ci run and we can run as many tests against it as we want.

So now some thoughts on getting the time down.

1. You are building the node with `--release`, but running the tests without it. I'm pretty sure that means that a brand new build from scratch is required. What if we do both with `--release` or both without?
2. Speaking of `--release` I know that is pretty common in Substrate ecosystem, but I wonder if we actually need it. a non release build will be faster of course.
3. If we can't get the node build to be an incremental build and it really does need to be a fresh build, then we definitely need to make it concurrent with the test job.
4. You mentioned that caching doesn't seem to help. I've noticed caching to be kind of hit or miss too. I'm always happy to get a cache hit, andit does happen sometimes, but I've come to never expect a hit.

Overall good work, and I'm happy to have this.

muraca:
1. I just did `cargo clean`, then `cargo clippy --release`, and later `cargo build --release`. The build started from scratch.
2. Fine for me to remove the release. I ran again the previous steps without it and everything is faster.
3. I think that we can just make it parallel and hope that 4. caches hits as much as possible!

muraca:
> Also notice that the marginal cost to add more e2e tests once this goes in is very small.

Once we have some script that does more tests, we can run that instead of `./target/debug/tuxedo-template-wallet --endpoint http://localhost:9944`

JoshOrndorff (approved):

PR #114, opened 2023-10-09
End of block inherents

JoshOrndorff:
This is an alternative approach to #105 and explores block rewards in a more utxo-native way. This will build on top of #100 and therefore I won't pursue it too much further until that lands.

The general idea is that block authors will have a chance to insert some additional inherent transactions at the end of the block. And in the case of tips specifically, this inherent will consume all of the Tip outputs from the block and reward the block author with a new UTXO.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8300,6 +8300,7 @@ dependencies = [
  "sp-api",
  "sp-core",
  "sp-debug-derive",
+ "sp-inherents",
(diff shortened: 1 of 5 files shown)


JoshOrndorff:
This is stale and there wasn't much here to begin with. We can pick these commits or just take inspiration whenever we get around to end-of-block inherents.

PR #123, opened 2023-10-13
GenesisBuilder with extrinsics execution

### LINKED ISSUE #107: Custom `GenesisBlockBuilder`
Related to my question on the [Substrate Stack Exchange: Extrinsics in Genesis Block](https://substrate.stackexchange.com/questions/10105/extrinsics-in-genesis-block).

Substrate chains by default do not have any extrinsics in the genesis block. This is fine in many cases, but we desire to have extrinsics in the genesis block. The most concrete use case I have in mind is putting a timestamp there. That would allow us to remove the first block hack introduced in #100. It would also be useful for initial token distributions and other usecases as explained in the stack exchange question.

To achieve this we need to implement the [`GeneisBlockBuilder` trait](https://paritytech.github.io/polkadot-sdk/master/sc_chain_spec/trait.BuildGenesisBlock.html). The defacto standard implementation in Substrate chains is the [`GenesisBlockBuilder` struct](https://paritytech.github.io/polkadot-sdk/master/sc_chain_spec/struct.GenesisBlockBuilder.html) and I propose we take inspiration from it but just add some genesis extrinsics.

Just braindumping, the implementatin might look something like this?

```rust
/// A Genesis block builder similar to the default in sc_chain_spec, but this one
/// allows extrinsics in the genesis block
pub struct TuxedoGenesisBlockBuilder {
  // Fields TBD
}

// Inspired by https://paritytech.github.io/polkadot-sdk/master/src/sc_chain_spec/genesis.rs.html#124
impl<Block: BlockT, B: Backend<Block>, E: RuntimeVersionOf> BuildGenesisBlock<Block>
	for TuxedoGenesisBlockBuilder<Block, B, E>
{
	type BlockImportOperation = <B as Backend<Block>>::BlockImportOperation;

	fn build_genesis_block(self) -> sp_blockchain::Result<(Block, Self::BlockImportOperation)> {
		let Self { genesis_pre_state, genesis_extrinsics, commit_genesis_state, backend, executor, _phantom } = self;

		let genesis_state_version = resolve_state_version_from_wasm(&genesis_storage, &executor)?;
		let mut op = backend.begin_operation()?;
		let mut genesis_block = construct_genesis_block::<Block>(state_root, genesis_state_version);

		// TODO push extrinsics to block
		// TODO calculate the new post-state. Maybe we can actually 
		// execute the transactions somehow; that would be ideal
		// Or maybe we can just manually construct the correct output state.

		let state_root =
			op.set_genesis_state(genesis_storage, commit_genesis_state, genesis_state_version)?;

		Ok((genesis_block, op))
	}
}
```

Perhaps we could also clean up the in-runtime aspects of the broader genesis by only allowing extrinsics to create state. Like when you fill in the genesis config in the chain spec, you actually just fill in some json like this.

```
genesis: {
  code: ...,
  timestamp: ...,
  genesis_extrinsics: [Mint(100, Alice), Mint(50, Bob)]
}
```

muraca:
I hope this will be enough to fix #107 

- [x] A custom TuxedoGenesisBuilder 
- [x] Use TuxedoGenesisBlockBuilder to initialize the first block
- [ ] Pass some extrinsics to TuxedoGenesisBlockBuilder
- [ ] Run and apply extrinsics in TuxedoGenesisBlockBuilder 
- [ ] Remove the special case for initializing timestamp in the first block

### DIFF
--- a/node/src/service.rs
+++ b/node/src/service.rs
@@ -1,6 +1,6 @@
 //! Service and ServiceFactory implementation. Specialized wrapper over substrate service.
 
-use crate::rpc;
+use crate::{rpc, tuxedo_genesis_builder::TuxedoGenesisBlockBuilder};
 use node_template_runtime::{self, opaque::Block, RuntimeApi};
 use sc_client_api::BlockBackend;
 use sc_consensus_aura::{ImportQueueParams, SlotProportion, StartAuraParams};
@@ -10,6 +10,7 @@ use sc_service::{error::Error as ServiceError, Configuration, TaskManager, WarpS
 use sc_telemetry::{Telemetry, TelemetryWorker};
 use sc_transaction_pool_api::OffchainTransactionPoolFactory;
 use sp_consensus_aura::sr25519::AuthorityPair as AuraPair;
+use sp_runtime::traits::Block as BlockT;
 use std::{sync::Arc, time::Duration};
 
 // Our native executor instance.
@@ -72,11 +73,23 @@ pub fn new_partial(
 
     let executor = sc_service::new_native_or_wasm_executor(config);
 
+    let extrinsics: Vec<<Block as BlockT>::Extrinsic> = vec![];
+    let backend = sc_service::new_db_backend(config.db_config())?;
+    let genesis_block_builder = TuxedoGenesisBlockBuilder::new(
+        config.chain_spec.as_storage_builder(),
+        !config.no_genesis(),
+        backend.clone(),
+        executor.clone(),
+        extrinsics,
+    )?;
+
     let (client, backend, keystore_container, task_manager) =
-        sc_service::new_full_parts::<Block, RuntimeApi, _>(
+        sc_service::new_full_parts_with_genesis_builder::<Block, RuntimeApi, _, _>(
             config,
             telemetry.as_ref().map(|(_, telemetry)| telemetry.handle()),
             executor,
+            backend,
+            genesis_block_builder,
         )?;
     let client = Arc::new(client);
 
--- a/node/src/service.rs
+++ b/node/src/service.rs
@@ -2,6 +2,7 @@
 
 use crate::{rpc, tuxedo_genesis_builder::TuxedoGenesisBlockBuilder};
 use node_template_runtime::{self, opaque::Block, RuntimeApi};
+use parity_scale_codec::Encode;
 use sc_client_api::BlockBackend;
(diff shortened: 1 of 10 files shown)


JoshOrndorff:
Yeah, I have one major suggestion that might help. We don't want to hardcode the genesis extrinsics in the service file. Rather we want the runtime to create them for us. For the first iteration it will just create the inherents, and later maybe we can add a way for runtime developers to specify some too. But the bit point is that the runtime should create the genesis extrinsics. We may need to provide a dedicated runtime API for that, but for starters you might be able to abuse the existing create_inherents API.

muraca:
I will have to build this on top of #100, so let's go on with that first.

muraca:
A huge problem I'm facing right now is that during genesis we can't access storage from sp_io, so from our Executive we can't use functions like `block_number` or `apply_tuxedo_transaction`.

node/src/service.rs
```diff
@@ -71,12 +75,44 @@ pub fn new_partial(
...
+    let timestamp = std::time::SystemTime::now()
+        .duration_since(std::time::SystemTime::UNIX_EPOCH)
+        .expect("Time is always after UNIX_EPOCH; qed")
+        .as_millis() as u64;
+
+    inherent_data
+        .put_data(sp_timestamp::INHERENT_IDENTIFIER, &timestamp)
+        .expect("Put timestamp inherent data");
+
+    let extrinsics: Vec<<Block as BlockT>::Extrinsic> =
```
JoshOrndorff, line 89: This is not a clean way to "cross the wasm boudary". The code you are writing here is on the client side. And you are calling `create_inherents` which is in the runtime. These kinds of calls should ideally go through a runtime api.

JoshOrndorff: As we discussed in element today, this is harder than I initially thought. It seems like genesis building may still be tied to the native runtime. I've asked about this in https://substrate.stackexchange.com/a/10128/372

So I'm okay with you calling directly into the runtime. But still, creating the extrinsics should not happen in the service file. The should happen in the genesis block builder implementation.

JoshOrndorff:
To capture what we agreed on in element. We will use this technique to put extrinsics into the genesis block, but we will not execute them. We will simply define any genesis transaction as valid. Instead of executing them, we just put all the outputs directly into genesis storage.

Optionally, we could make sure there are no inputs or peeks in genesis transactions, but I don't think it adds much value.

PR #124, opened 2023-10-13
make CI delete old comments

muraca:
While reviewing #100 I noticed that aforesaid feature silently introduced in #113 does not work.
I hope switching to this other fork of the action will fix it, without introducing other problems.


JoshOrndorff:
Awesome! And since the bot deleted its own comment, we don't even have one of these ugly log items.

image

PR #126, opened 2023-10-13
Remove panics from service's `create_inherent_data_providers` function

### LINKED ISSUE #125: Replace `expect` with Error in `node/src/service.rs`
I was trying earlier to make this return an error when needed, instead of panicking. I wanted to do someting like:
```rust
let parent_block = client_for_cidp
    .clone()
    .block(parent_hash)?
    .ok_or(sp_blockchain::Error::UnknownBlock(parent_hash))?
    .block;
```
There was some issues with the closure being async, and I'm not familiar with that, so please try to make this work without spending too much time on it, otherwise it's fine to leave it like this.

_Originally posted by @muraca in https://github.com/Off-Narrative-Labs/Tuxedo/pull/100#discussion_r1358478110_

JoshOrndorff:
Closes #125 

Wow, we really can handle these errors gracefully. Thanks for the encouragement @muraca.

This PR removes some potential panics from the node's service. Specifically the part where we are creating the inherent data providers used during the authoring process.

This is complex because it is a function that returns a function that returns a result. In the outer function we have access to the client, so we absolutely must read the parent block from the db at that point. But we can't return a result there, we just have to always return the function, so instead I was just panicing.

The trick is to query the database in the outer function, but defer all of the error handling until the inner function which can return a result.

### DIFF
--- a/node/src/service.rs
+++ b/node/src/service.rs
@@ -244,14 +244,12 @@ pub fn new_full(config: Configuration) -> Result<TaskManager, ServiceError> {
                 block_import,
                 proposer_factory,
                 create_inherent_data_providers: move |parent_hash, ()| {
-                    let parent_block = client_for_cidp
-                        .clone()
-                        .block(parent_hash)
-                        .expect("Block backend should not error.")
-                        .expect("Parent block should be found in database for authoring to work.")
-                        .block;
+                    let maybe_parent_block = client_for_cidp.clone().block(parent_hash);
 
                     async move {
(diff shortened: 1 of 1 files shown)


muraca:
https://github.com/Off-Narrative-Labs/Tuxedo/blob/c5788f544ce59d1e6d525c12d5f61c7619bcd26f/node/src/service.rs#L165

I know that this line is taken from Substrate Node Template, but since we're here let's try to remove it as well please 🙏🏾

muraca (approved):

PR #127, opened 2023-10-14
Genesis Transactions

### LINKED ISSUE #107: Custom `GenesisBlockBuilder`

muraca:
This PR introduces genesis transactions, which are neither validated nor executed.
They are stored in the genesis block, and their outputs are saved in storage as well.
Resolves #107 and blocks #128

Changelog:
- Added `std`-gated function `InherentHooks::genesis_transactions` to generate genesis inherents.
- `GenesisConfig`'s only parameter is now a `Vec<Transaction>`.
- Created file `tuxedo-core/src/genesis.rs` which contains:
	- A custom `TuxedoGenesisBlockBuilder` based on the one I implemented in #123, which reads the extrinsics list from storage key `EXTRINSICS_KEY`, to include them in the genesis block.
	- The `std`-gated function `assimilate_storage`, which:
	    - stores the whole `genesis_transactions` list after the key `EXTRINSICS_KEY`;
	    - checks that the `Transaction`s have no inputs or peeks;
	    - the `Output`s of the `genesis_transactions` are stored based on the `Transaction` that created them.
- The function `<TuxedoGenesisConfig as BuildStorage>::assimilate_storage`:
	- builds the actual `genesis_transactions` list appending at `InherentHooks::genesis_transactions` its  `TuxedoGenesisConfig.genesis_transactions`, since the inherents should appear at the beginning of the block;
	- calls `genesis.rs/assimilate_storage`;
	- is later called by the `TuxedoGenesisBlockBuilder` using the trait `BuildStorage`.
- Updated wallet's initialization and synchronization to work with the new genesis blocks.


wardrobe/timestamp/src/lib.rs
```diff
@@ -321,6 +312,19 @@ impl<V: Verifier + From<UpForGrabs>, T: TimestampConfig + 'static> TuxedoInheren
                 .expect("Should be able to push some error");
         }
     }
+
+    #[cfg(feature = "std")]
+    fn genesis_utxos() -> Vec<Output<V>> {
+        let time = std::time::SystemTime::now()
+            .duration_since(std::time::SystemTime::UNIX_EPOCH)
+            .expect("Time is always after UNIX_EPOCH; qed")
+            .as_millis() as u64;
```
JoshOrndorff, line 321: I've been wondering why this was all feature-gated to `std`. Now I see it is because you are directly querying the time. This design is not sustainable because the native runtime is almost removed from Substrate. The way we typically pass the "local time" into the runtime is through inherent data. It comes into both `inherent_extrinsics` and `check_inherents` this way. I recommend we do the same thing here. (Or for a quick0n0dirty prototype, just take the local time as a parameter directly.)

JoshOrndorff (commented):
I pushed a commit to make this compile. It was very close already; nice work on that.

When I run the development node, it panics when trying to author block 1. This is what I expected. The problem is that when your node is authoring block 1. It needs to know what the UtxoRef for the previous timestamp was. The way that the authoring process finds out the previous timestamp is to scrape it from the previous block's extrinsics.

Then when the inherent is executed on chain, the previous inherent needs to be available in storage to be peeked at. You've solved this second problem, but not the first.


```
$ ./target/release/node-template --dev
2023-10-16 09:35:13 Tuxedo Template Node    
2023-10-16 09:35:13 ✌️  version 4.0.0-dev-0c3ebbdbb64    
2023-10-16 09:35:13 ❤️  by , 2023-2023    
2023-10-16 09:35:13 📋 Chain specification: Development    
2023-10-16 09:35:13 🏷  Node name: big-print-2343    
2023-10-16 09:35:13 👤 Role: AUTHORITY    
2023-10-16 09:35:13 💾 Database: RocksDb at /tmp/substrateWi2b5G/chains/dev/db/full    
2023-10-16 09:35:14 🔨 Initializing Genesis block/state (state: 0xd9c0…cbe5, header-hash: 0xd283…4622)    
2023-10-16 09:35:14 👴 Loading GRANDPA authority set from genesis on what appears to be first startup.    
2023-10-16 09:35:14 Using default protocol ID "sup" because none is configured in the chain specs    
2023-10-16 09:35:14 🏷  Local node identity is: 12D3KooWGH9zrgw8P9y9PnXbrMm9cza53TNyT5xTdVSPKbdSsWiZ    
2023-10-16 09:35:14 💻 Operating system: linux    
2023-10-16 09:35:14 💻 CPU architecture: x86_64    
2023-10-16 09:35:14 💻 Target environment: gnu    
2023-10-16 09:35:14 💻 CPU: 13th Gen Intel(R) Core(TM) i9-13900HX    
2023-10-16 09:35:14 💻 CPU cores: 24    
2023-10-16 09:35:14 💻 Memory: 64135MB    
2023-10-16 09:35:14 💻 Kernel: 6.5.4-76060504-generic    
2023-10-16 09:35:14 💻 Linux distribution: Pop!_OS 22.04 LTS    
2023-10-16 09:35:14 💻 Virtual machine: no    
2023-10-16 09:35:14 📦 Highest known block at #0    
2023-10-16 09:35:14 Running JSON-RPC server: addr=127.0.0.1:9944, allowed origins=["*"]    
2023-10-16 09:35:14 〽️ Prometheus exporter started at 127.0.0.1:9615    
2023-10-16 09:35:15 🙌 Starting consensus session on top of parent 0xd2838227064ca9e8bcd213ecea93b9033fdaf24be2860a38783d5d0a2c784622    
2023-10-16 09:35:15 panicked at 'Attemping to construct timestamp inherent with no previous inherent.', /home/joshy/ProgrammingProjects/tuxedo/wardrobe/timestamp/src/lib.rs:251:13    
2023-10-16 09:35:15 Proposing failed: Import failed: Execution failed: Execution aborted due to trap: wasm trap: wasm `unreachable` instruction executed
WASM backtrace:
error while executing at wasm backtrace:
    0: 0x695c - <unknown>!rust_begin_unwind
    1: 0x1a7e - <unknown>!core::panicking::panic_fmt::h7859ddfa497c0e61
    2: 0x2590b - <unknown>!<timestamp::SetTimestamp<T> as tuxedo_core::inherents::TuxedoInherent<V,timestamp::SetTimestamp<T>>>::create_inherent::h6e481c843df9c309
    3: 0x2241e - <unknown>!<tuxedo_core::inherents::TuxedoInherentAdapter<T> as tuxedo_core::inherents::InherentInternal<V,C>>::create_inherents::h5444997e0e25e64a
    4: 0x1822b - <unknown>!<tuxedo_template_runtime::OuterConstraintCheckerInherentHooks as tuxedo_core::inherents::InherentInternal<tuxedo_template_runtime::OuterVerifier,tuxedo_template_runtime::OuterConstraintChecker>>::create_inherents::h0bed55d9ba3f3695
    5: 0x1bf91 - <unknown>!tuxedo_core::executive::Executive<B,V,C>::inherent_extrinsics::h2ba265801b74798f
    6: 0x29a11 - <unknown>!BlockBuilder_inherent_extrinsics    
2023-10-16 09:35:18 🙌 Starting consensus session on top of parent 0xd2838227064ca9e8bcd213ecea93b9033fdaf24be2860a38783d5d0a2c784622    
2023-10-16 09:35:18 panicked at 'Attemping to construct timestamp inherent with no previous inherent.', /home/joshy/ProgrammingProjects/tuxedo/wardrobe/timestamp/src/lib.rs:251:13    
2023-10-16 09:35:18 Proposing failed: Import failed: Execution failed: Execution aborted due to trap: wasm trap: wasm `unreachable` instruction executed
WASM backtrace:
error while executing at wasm backtrace:
    0: 0x695c - <unknown>!rust_begin_unwind
    1: 0x1a7e - <unknown>!core::panicking::panic_fmt::h7859ddfa497c0e61
    2: 0x2590b - <unknown>!<timestamp::SetTimestamp<T> as tuxedo_core::inherents::TuxedoInherent<V,timestamp::SetTimestamp<T>>>::create_inherent::h6e481c843df9c309
    3: 0x2241e - <unknown>!<tuxedo_core::inherents::TuxedoInherentAdapter<T> as tuxedo_core::inherents::InherentInternal<V,C>>::create_inherents::h5444997e0e25e64a
    4: 0x1822b - <unknown>!<tuxedo_template_runtime::OuterConstraintCheckerInherentHooks as tuxedo_core::inherents::InherentInternal<tuxedo_template_runtime::OuterVerifier,tuxedo_template_runtime::OuterConstraintChecker>>::create_inherents::h0bed55d9ba3f3695
    5: 0x1bf91 - <unknown>!tuxedo_core::executive::Executive<B,V,C>::inherent_extrinsics::h2ba265801b74798f
    6: 0x29a11 - <unknown>!BlockBuilder_inherent_extrinsics    
2023-10-16 09:35:19 💤 Idle (0 peers), best: #0 (0xd283…4622), finalized #0 (0xd283…4622), ⬇ 0 ⬆ 0    
```

muraca:
Quick recap of what happens:
- The function initially called `InherentHooks::genesis_utxos` was changed to `InherentHooks::genesis_transactions`.
- `GenesisConfig`'s only parameter is now a `Vec<Transaction>`.
- In the `std`-gated function `<GenesisConfig as BuildStorage>::assimilate_storage`, which is later called by the `GenesisBuilder`:
  - the `genesis_transactions` is appended at the result of `InherentHooks::genesis_transactions`, since the inherents should appear at the beginning of the block;
  - the whole `genesis_transactions` list is stored after the key `EXTRINSICS_KEY`; 
  - the `Output`s are stored based on the `Transaction` that created them.
- A custom `TuxedoGenesisBlockBuilder` based on the one I implemented in #123 is used instead of the default one, which:
  - is currently stored in crate `node`, but I think the `tuxedo-core` would be a better fit;
  - reads the extrinsics list from storage, to include them in the genesis block; 
  - currently does not clear the storage item for the key `EXTRINSICS_KEY`; a temporary fix was put in the `Executive` code [here](https://github.com/Off-Narrative-Labs/Tuxedo/blob/025b4d336443204e7ee70f7fd22c3006056caf65/tuxedo-core/src/executive.rs#L245), to clear the storage before adding the extrinsics for the next block.

muraca:
@JoshOrndorff I think the `init_from_genesis` function in the wallet is superfluous, as we can synchronize using the same function we use for all the other blocks. The wallet test fails because this function fails, and if you remove that call, everything works as expected. 
I'm not 100% familiar with the wallet yet, so I'll wait for your response on that.

tuxedo-core/src/executive.rs
```diff
@@ -239,6 +239,10 @@ impl<
         // Store the transient partial header for updating at the end of the block.
         // This will be removed from storage before the end of the block.
         sp_io::storage::set(HEADER_KEY, &header.encode());
+
+        // Clear the extrinsics list from the previous block.
+        // TODO Remove: currently this is needed because we still need to find a way to clear it in block 0.
+        sp_io::storage::clear(EXTRINSIC_KEY);
```
JoshOrndorff, line 245: I wasn't expecting to need this. Extrinsics only get noted under that key when they are executed. But the genesis extrinsics (and genesis block in general) are not executed.

My understanding is that when a node starts authoring block 1 they will start with a storage that only contains whatever the genesis builder put there and nothing else.

tuxedo-template-runtime/src/lib.rs
```diff
@@ -119,34 +119,34 @@ impl Default for GenesisConfig {
...
+
+        GenesisConfig {
+            genesis_transactions,
         }
 
-        // TODO: Initial UTXO for Kitties
+        // TODO: Initial Transactions for Kitties
 
-        // TODO: Initial UTXO for Existence
+        // TODO: Initial Transactions for Existence
```
JoshOrndorff, line 149: Seems like a good time to resolve these. I think any example from any other piece would be sufficient. So maybe add a single PoE claim to genesis block, or a single amoeba creation to the genesis block.

tuxedo-template-runtime/src/lib.rs
```diff
@@ -159,17 +159,26 @@ impl BuildStorage for GenesisConfig {
...
-                index: index as u32,
-            };
-            storage.top.insert(output_ref.encode(), utxo.encode());
+        let mut genesis_transactions: Vec<Transaction> =
+            OuterConstraintCheckerInherentHooks::genesis_transactions();
+        genesis_transactions.extend(self.genesis_transactions.clone());
+
+        storage
+            .top
+            .insert(EXTRINSIC_KEY.to_vec(), genesis_transactions.encode());
```
JoshOrndorff, line 171: The comment at the beginning of this function needs revised now that we are doing interesting things here.

tuxedo-core/src/executive.rs
JoshOrndorff: Nvm, I didn't understand the storage use yet.

muraca: It it used by `assimilate_storage` to send the transactions to the `GenesisBlockBuilder` https://github.com/Off-Narrative-Labs/Tuxedo/blob/fba140db0d4c43fc5441dc6b7167a7807f622e9a/tuxedo-template-runtime/src/lib.rs#L189 
But `GenesisBlockBuilder` can't mutate the storage

JoshOrndorff:
I just found out there is some ongoing issue in Substrate about genesis block builder and genesis config and chain spec and how to do it after the native runtime is deprecated.

https://github.com/paritytech/polkadot-sdk/issues/25

muraca:
@JoshOrndorff 

> I just found out there is some ongoing issue in Substrate about genesis block builder and genesis config and chain spec and how to do it after the native runtime is deprecated.
> 
> [paritytech/polkadot-sdk#25](https://github.com/paritytech/polkadot-sdk/issues/25)

I'd say let's conclude what we have going on now, and later adhere to the new native runtime free standards when everything is finished and well documented.

JoshOrndorff (approved):
Nice work!

JoshOrndorff:
> Partially resolves https://github.com/Off-Narrative-Labs/Tuxedo/issues/107, we should open another issue to try to execute the transactions later on.

I'm not sure executing them adds much value. This happens only at chain launch time, and I think that looking carefully at the genesis block (especially now that it explicitly contains all genesis state) is the duty of someone wanting to use the chain. Basically I'm arguing that the genesis block is like a summary of the prehistory and it's okay if the semantics are squishy or manually enforced for this one-time process to launch the chain.

So IMO this resolves #107 fully.

PR #128, opened 2023-10-16
beautify `TuxedoGenesisConfig`

### LINKED ISSUE #23: Genesis config
The [current genesis config](https://github.com/Off-Narrative-Labs/Tuxedo/blob/7f1dabe37ce8621deb9df47d102ea775bbe2f763/frameless-runtime/src/lib.rs#L107-L109) could be a little nicer.

```rust
https://github.com/Off-Narrative-Labs/Tuxedo/blob/7f1dabe37ce8621deb9df47d102ea775bbe2f763/frameless-runtime/src/lib.rs#L107-L109
```

In the UTXO model the state is (basically) just the utxo set. In this sense, the current design is perfect with its single Vec of outputs. But there are still a few things that may be desired.

1. Including a money in the default genesis config is unusual. The default should be empty and the output that goes to `SHAWN_PUB_KEY` should be expressed in one or more of the specs in `chain_spec.rs`.
2. We should figure out the best-practice way of expressing the genesis utxo set in the `chain_spec.rs` file and provide helper functions or builder pattern or something if necessary. Off the top of my head, maybe something like.

```rust
let genesis_utxo_set = vec![
  (Coin(5), SigCheck(ALICE_PUB_KEY)).into(),
  (Amoeba{ generation: 5, ..Default::default() }, UpForGrabs).into()
];
```

muraca:
resolves #23 
New syntax is:
https://github.com/Off-Narrative-Labs/Tuxedo/blob/02a463fbd20369f24fe8f0f3bb53cd896d2779cd/tuxedo-template-runtime/src/lib.rs#L121-L128

This is done by leveraging functions written ad-hoc.
https://github.com/Off-Narrative-Labs/Tuxedo/blob/02a463fbd20369f24fe8f0f3bb53cd896d2779cd/wardrobe/money/src/lib.rs#L80-L93

`TuxedoGenesisConfig` has been moved to `tuxedo-core/src/genesis.rs`, while the runtime type `RuntimeGenesisConfig`, the tests and the function that replaces the Default trait implementation have been moved to `tuxedo-template-runtime/src/genesis.rs`.

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -2,7 +2,8 @@ use hex_literal::hex;
 use node_template_runtime::{
     kitties::{KittyData, Parent},
     money::Coin,
-    OuterConstraintChecker, OuterConstraintCheckerInherentHooks, OuterVerifier, WASM_BINARY,
+    OuterConstraintChecker, OuterConstraintCheckerInherentHooks, OuterVerifier,
+    RuntimeGenesisConfig, WASM_BINARY,
 };
 use sc_service::ChainType;
 use tuxedo_core::{
@@ -15,8 +16,7 @@ use tuxedo_core::{
 // const STAGING_TELEMETRY_URL: &str = "wss://telemetry.polkadot.io/submit/";
 
 /// Specialized `ChainSpec`. This is a specialization of the general Substrate ChainSpec type.
-pub type ChainSpec =
-    sc_service::GenericChainSpec<TuxedoGenesisConfig<OuterVerifier, OuterConstraintChecker>>;
+pub type ChainSpec = sc_service::GenericChainSpec<RuntimeGenesisConfig>;
 
 // /// Generate a crypto pair from seed.
 // pub fn get_from_seed<TPublic: Public>(seed: &str) -> <TPublic::Pair as Pair>::Public {
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -1,16 +1,5 @@
-use hex_literal::hex;
-use node_template_runtime::{
-    kitties::{KittyData, Parent},
-    money::Coin,
-    OuterConstraintChecker, OuterConstraintCheckerInherentHooks, OuterVerifier,
-    RuntimeGenesisConfig, WASM_BINARY,
-};
+use node_template_runtime::genesis::*;
 use sc_service::ChainType;
-use tuxedo_core::{
-    genesis::TuxedoGenesisConfig,
-    inherents::InherentInternal,
-    verifier::{SigCheck, ThresholdMultiSignature, UpForGrabs},
-};
 
 // The URL for the telemetry server.
 // const STAGING_TELEMETRY_URL: &str = "wss://telemetry.polkadot.io/submit/";
@@ -39,35 +28,6 @@ pub type ChainSpec = sc_service::GenericChainSpec<RuntimeGenesisConfig>;
 // pub fn authority_keys_from_seed(s: &str) -> (AuraId, GrandpaId) {
 // 	(get_from_seed::<AuraId>(s), get_from_seed::<GrandpaId>(s))
 // }
-
-const SHAWN_PUB_KEY_BYTES: [u8; 32] =
-    hex!("d2bf4b844dfefd6772a8843e669f943408966a977e3ae2af1dd78e0f55f4df67");
-const ANDREW_PUB_KEY_BYTES: [u8; 32] =
-    hex!("baa81e58b1b4d053c2e86d93045765036f9d265c7dfe8b9693bbc2c0f048d93a");
-
-fn development_genesis_config() -> TuxedoGenesisConfig<OuterVerifier, OuterConstraintChecker> {
-    let signatories = vec![SHAWN_PUB_KEY_BYTES.into(), ANDREW_PUB_KEY_BYTES.into()];
-
-    // The inherents are computed using the appropriate method, and placed before the extrinsics.
-    let mut genesis_transactions = OuterConstraintCheckerInherentHooks::genesis_transactions();
-
-    genesis_transactions.extend([
-        // Money Transactions
-        Coin::<0>::mint(100, SigCheck::new(SHAWN_PUB_KEY_BYTES)),
-        Coin::<0>::mint(100, ThresholdMultiSignature::new(1, signatories)),
-        // Kitty Transactions
-        KittyData::mint(Parent::mom(), b"mother", UpForGrabs),
-        KittyData::mint(Parent::dad(), b"father", UpForGrabs),
-        // TODO: Initial Transactions for Existence
-    ]);
-
-    TuxedoGenesisConfig::new(
-        WASM_BINARY
-            .expect("Runtime WASM binary must exist.")
-            .to_vec(),
-        genesis_transactions,
-    )
-}
 
 pub fn development_config() -> Result<ChainSpec, String> {
     Ok(ChainSpec::from_genesis(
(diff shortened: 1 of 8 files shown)


tuxedo-template-runtime/src/lib.rs
```diff
@@ -113,40 +112,25 @@ pub struct GenesisConfig {
...
     fn default() -> Self {
         use hex_literal::hex;
+        use money::Coin;
 
         const SHAWN_PUB_KEY_BYTES: [u8; 32] =
             hex!("d2bf4b844dfefd6772a8843e669f943408966a977e3ae2af1dd78e0f55f4df67");
         const ANDREW_PUB_KEY_BYTES: [u8; 32] =
             hex!("baa81e58b1b4d053c2e86d93045765036f9d265c7dfe8b9693bbc2c0f048d93a");
 
+        let signers = vec![SHAWN_PUB_KEY_BYTES.into(), ANDREW_PUB_KEY_BYTES.into()];
```
JoshOrndorff, line 122:
```suggestion
        let genesis_multisig_signatories = vec![SHAWN_PUB_KEY_BYTES.into(), ANDREW_PUB_KEY_BYTES.into()];
```

JoshOrndorff (approved):

JoshOrndorff:
I approved this earlier because it is certainly a big improvement over what we have previously.

But with the genesis extrinsics shaping up in #123 and #127, I wonder if we should forbid runtime developers from manually creating outputs directly, and rather require them to only create genesis extrinsics, which will specify the genesis utxos as their outputs.

I think that requiring genesis transactions encourages the genesis block authors to communicate some semantics to the posterity of their chain. And it also makes the genesis block read more like a creation myth.

muraca:
> I wonder if we should forbid runtime developers from manually creating outputs directly, and rather require them to only create genesis extrinsics, which will specify the genesis utxos as their outputs.

@JoshOrndorff I might agree on the fact that we should put Transactions in the GenesisBuilder, rather than Outputs, as we currently do. 
However, I think it's a good idea to use this syntax simplify the process for who wants to bootstrap a chain, since said Transactions should not consume any input or use peeks.

JoshOrndorff:
Agreed

JoshOrndorff:
Perhaps this is a good time to address some technical debt that Andrew and I introduced early on in order to keep prototyping quickly.

Ideally, the genesis extrinsics for a given chain would be specified in the chain spec. If you look at the Substrate node template or other FRAME chains, they do this. All the genesis state is specified directly in the chain spec. So if I want to launch a node template with my own initial balances or staking authorities, I only edit the chain spec, no need to re-compile anything. This is what we want in Tuxedo too (except it will be genesis transactions, not genesis state.)

The  current design where we hard-code it all in the `impl Default for GenesisBuilder` is the hack / tech debt. Really the default should be something simple like only storing the wasm runtime, and not having any initial funds.

Regarding the piece-specific helper function, I totally support that idea. Those functions could be called form the chain spec.

node/src/chain_spec.rs
```diff
@@ -1,11 +1,22 @@
...
+    verifier::{SigCheck, ThresholdMultiSignature, UpForGrabs},
+};
 
 // The URL for the telemetry server.
 // const STAGING_TELEMETRY_URL: &str = "wss://telemetry.polkadot.io/submit/";
 
 /// Specialized `ChainSpec`. This is a specialization of the general Substrate ChainSpec type.
-pub type ChainSpec = sc_service::GenericChainSpec<TuxedoGenesisConfig>;
+pub type ChainSpec =
+    sc_service::GenericChainSpec<TuxedoGenesisConfig<OuterVerifier, OuterConstraintChecker>>;
```
JoshOrndorff, line 19: Consider making a public type alias `type GenesisConfig = TuxedoGenesisConfig<OuterV, OuterC>;` in the runtime. We do this also for Transaction, Block, and I think other Tuxedo core types.

Basically the idea is that the Runtime provides the fully concrete types.

node/src/chain_spec.rs
```diff
@@ -29,14 +40,44 @@ pub type ChainSpec = sc_service::GenericChainSpec<TuxedoGenesisConfig>;
 // 	(get_from_seed::<AuraId>(s), get_from_seed::<GrandpaId>(s))
 // }
```
muraca, line 41: @JoshOrndorff can this commented part be removed?

muraca: https://github.com/Off-Narrative-Labs/Tuxedo/blob/52b0b49e0930e10e71b3cdddf0f519cb854babe1/node/src/chain_spec.rs#L21-L41

muraca: This one as well:
https://github.com/Off-Narrative-Labs/Tuxedo/blob/52b0b49e0930e10e71b3cdddf0f519cb854babe1/node/src/chain_spec.rs#L15-L16

JoshOrndorff: The telemetry address should stay. It is [copied directly](https://github.com/paritytech/polkadot-sdk/blob/6b7be115fd523057d01cea46e9680a62a0b08d97/substrate/bin/node-template/node/src/chain_spec.rs#L9) from the upstream node template (commented and all).

The reason they have it there is because production chains will want it enabled and it is nice to have it readily available. The reason it is commented is because Parity doesn't want every learner who spins up a node template spamming their telemetry server.

Basically, one of my goals is to keep the diff between our node template's client side and the upstream node template's client side as small as practical.

muraca: Got it. Let me know about the other part then.

JoshOrndorff: For the main question about get_from_seed and stuff. I basically just haven't fully understood this stuff. I don't know if we will be useful when we start having proper authority selection in the runtime, etc. I don't think we should remove it in this PR at least. I'll make an issue describing some ideas I have in mind about authority selection.

muraca: Let me know if we can merge this, so I can move on with #132 as well

JoshOrndorff (commented):
I'm pretty happy with the design here. There are just a couple last things that still seems a little out-of-place to me.

One is the tests in the runtime. Are they just creating some random genesis config and then testing that it was put in the storage correctly? Do they have much to do with the runtime itself?

I guess at minimum they should be moved to a tests.rs file in the runtime, but I wonder if they can just be deleted entirely. Now that the genesis config is better expressed.

Second is the function `development_genesis_config()`. I'm very happy that this is no longer in the `impl Default` block. It is already a huge improvement. But I wonder if moving it all the way to the node is too much. What if we put it in the runtime in a dedicated genesis.rs file. I can imagine that in the future we may want to add other similar functions. For example on that takes list of `endowed_addresses` or something. I followed this pattern of having a genesis module in the runtime back in the Substrate recipes days, and it worked well. Here are Five runtimes that all follow the pattern to see what I mean. https://github.com/JoshOrndorff/recipes/tree/master/runtimes

muraca:
> But I wonder if moving it all the way to the node is too much.  

I defaulted to that as it is what the substrate template does, with closures. 

> What if we put it in the runtime in a dedicated genesis.rs file.  

I see the value of this. The genesis should be tied to the runtime in my opinion as well. 

> I guess at minimum they should be moved to a tests.rs file in the runtime, but I wonder if they can just be deleted entirely. Now that the genesis config is better expressed.

I see the value of testing the genesis config works as expected, and this function is actually testing the `mint` functions implemented in this PR as well. 
I think the right place for these tests is in the `runtime/src/genesis.rs` along the aforementioned functions.

JoshOrndorff (approved):

PR #129, opened 2023-10-16
Adding dependency `cumulus-primitives-core` breaks runtime build

JoshOrndorff:
@bkchr I hope maybe you can help me understand this problem I'm having when trying to work with cumulus. My ultimate goal is to convert my runtime to be cumulus compatible, but I'm hitting a specific problem, and this branch is a minimal reproducible example.

For starters, notice that before this branch, I can build the runtime just fine with eg `cargo check -p tuxedo-template-runtime`.

All I've changed in this branch is to add a dependency from my runtime to the crate `cumulus-primitives-core`. I have not changed the actual code at all, just the `Cargo.toml` file.

With the  dependency added, when I run the same build command, the build fails complaining that one of my typed does not implement `Serialize`. I'm confused because [the type does derive Serialize](https://github.com/Off-Narrative-Labs/Tuxedo/blob/bad03b6e9b075858b4c0b4253a23b9a42c7f6fae/tuxedo-core/src/types.rs#L37) and the runtime compiles fine without this dependency.

And for a little more context, this `Transaction` type is the one that implements the `Extrinsic` trait and is used as the block's extrinsic type.

Thanks in advance for any advice.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -277,6 +277,20 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a4668cab20f66d8d020e1fbc0ebe47217433c1b6c8f2040faf858554e394ace6"
 
 [[package]]
+name = "aquamarine"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


JoshOrndorff:
This was related to enabling the serde feature on a few core crates. I've sorted it out more carefully in #131

PR #135, opened 2023-11-02
implement `--tmp` and `--dev` for wallet

### LINKED ISSUE #133: Wallet: Add command to clear all data
I'm lazy and I use the wallet in tests without specifying a custom data path. 
I need to manually go to my platform-specific folder and delete manually each time I need to restart the chain.
I think we should ad a command to clear all wallet data.
It should either have a long name, so people can't run it by mistake, or be conditionally compiled, if we only want to keep it for testing purposes.

muraca:
closes #133 

- default key Shawn is not included in the keystore anymore 
image1
- it is only included when the `--dev` arg is set
image2
- arg `--tmp` will clear the data folder at the end. `--dev` enables `--tmp` under the hood.
image3

I also improved some documentation here and there, mostly added `verbatim_doc_comment` to show the entire doc comment and support newlines.
image4

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,45 +27,45 @@ jobs:
           all: true
 
   test:
-      name: Test and code coverage
-      needs: [fmt, toml-sort]
-      runs-on: ubuntu-latest
-      permissions:
-          issues: write
-          pull-requests: write
-      steps:
-        - uses: actions/checkout@v4
-        - name: Install tooling
-          run: |
-            sudo apt-get install -y protobuf-compiler
-            protoc --version
-        - name: Install latest nextest release
-          uses: taiki-e/install-action@nextest
-        - name: Install cargo-llvm-cov
-          uses: taiki-e/install-action@cargo-llvm-cov
-        - name: Rust Cache
-          uses: Swatinem/rust-cache@v2
-          with:
-            cache-targets: true
-            cache-on-failure: true
-        - name: Run tests and print coverage data
-          run: cargo llvm-cov nextest --json --output-path lcov.json
-               --ignore-filename-regex "node/" --summary-only &&
-               echo "Lines coverage " && jq ".data[0].totals.lines.percent" lcov.json
-        # if the PR is on the same repo, the coverage data can be reported as a comment
-        - if: github.event_name == 'pull_request' &&
-              github.event.pull_request.head.repo.full_name == github.repository
-          name: Generate lcov report
-          run: cargo llvm-cov report --lcov --output-path lcov.info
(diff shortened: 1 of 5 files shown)


muraca:
@JoshOrndorff I think `data-path` might be refactored to `path`, I think `data` doesn't give any useful information.

JoshOrndorff (approved):
Overall great work. A couple little comments, but I trust you to address them (or not) as you see fit. Merge when ready.

1. I was thinking that Alice was the default/development key and SHAWN was only used in one or two places. Can you just double check that you got the right one.

2. What are the differences in the way the docs are rendered now? Ideally that would go in a separate PR. For example, in case there is some little bug in `--dev` and we want to revert it. We would not want to also revert the doc changes. But not a big deal in this case.
3. Looks like we upped the sleep time from 10s to 20s in the CI. Was 10 not long enough?

JoshOrndorff (changes requested):
Sorry, I missed something.

When we specify `--tmp` we need to generate a _new_ random path and use it for the run, and clean it up at the end. The current PR does not generate a new random path. It just relies on the existing methods to choose the path.

This is dangerous because someone could have a "real" wallet that they hold real funds in. And they don't want to be purging all of it's data. They may also do some development work and want to use `--tmp` for the dev work.  We would not want this person to simply add `--tmp` to the cli and accidentally purge their real data.

If you really like the existing semantics, I think a better flag name would be `--purge-when-done` or similar. But I like the idea to generate a random path better because that's exactly what Substrate does.

muraca:
> 1. I was thinking that Alice was the default/development key and SHAWN was only used in one or two places. Can you just double check that you got the right one.

Looks like Alice is never used in the wallet, Shawn is the default development key and it is also used in the template runtime. 

> 2. What are the differences in the way the docs are rendered now?

New lines are respected, allowing some longer descriptions to be displayed correctly.  
I get your point, but it's both a small change and something clearly stated in the PR description, I think if one ever had to revert the PR they would notice and maybe apply the changes elsewhere.

> 3. Looks like we upped the sleep time from 10s to 20s in the CI. Was 10 not long enough?

It was barely enough to kickstart the node, but not to produce any blocks. I should mention it in the changelog in the PR description as well.

> When we specify --tmp we need to generate a new random path and use it for the run 

I agree. I'll proceed with that.

muraca:
https://github.com/Off-Narrative-Labs/Tuxedo/blob/20efeb5b0efed016578e3937a0bba28f75757e18/wallet/src/cli.rs#L23

Is this comment still relevant?

JoshOrndorff (approved):
You’re right. That comment is outdated and should be removed.

PR #136, opened 2023-11-07
`rocksdb` feature flag

muraca:
Use ParityDB as the default database, and allow RocksDB to be used by compiling the node with the feature flag `rocksdb`.
This was done because building RocksDB is only a waste of time and resources if used for CI and tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -716,27 +716,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "bindgen"
-version = "0.65.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cfdf7b466f9a4903edc73f95d6d2bcd5baf8ae620638762244d3f60143643cc5"
-dependencies = [
- "bitflags 1.3.2",
- "cexpr",
- "clang-sys",
- "lazy_static",
- "lazycell",
- "peeking_take_while",
- "prettyplease 0.2.15",
- "proc-macro2",
- "quote",
- "regex",
- "rustc-hash",
- "shlex",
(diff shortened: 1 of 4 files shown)


JoshOrndorff:
My first thoughts were, "Is this really necessary, is parity db good enough, does Tuxedo need to be opinionated about the DB, etc".

But the compile time is a pretty compelling reason. How many hours have I wasted and CO2  emitted compiling `lib-rocks-db-sys`.

So in the end I actually do support this change, I just request that you document this a little bit better because this is a template and we don't want users getting confused and frustrated by things that are not so core to Tuxedo.

So for example, please provide a better PR description. And maybe also a comment in the `Cargo.toml` file where we disable the default features.

FYI, if the rock-db feature is enabled, the default DB is rocks DB. And it is a default feature. So in that sense, the default in the ecosystem is still to use Rocks DB.
https://github.com/paritytech/polkadot-sdk/blob/d347d68868a38841ba51f2e91cef945a7ea50dd4/substrate/client/cli/src/config.rs#L454-L463

muraca:
> FYI, if the rock-db feature is enabled, the default DB is rocks DB. And it is a default feature. So in that sense, the default in the ecosystem is still to use Rocks DB. 

Shall I use the same pattern? I thought it was an overkill, but if it's still the standard for Substrate then maybe it's the right thing to do.

JoshOrndorff (approved):
I also thought their way of determining the default was overkill. I wonder if it is for backwards compatibility for nodes that ahve been running since before parity db?

Anyway, I'm happy with the way you have it and look forward to the faster builds.

PR #130, opened 2023-10-16
Parachain support

JoshOrndorff:
This PR is the bulk of the effort to enable parachain support (including template runtime and node and dev service).

Some prerequisites include #100 and #131.
Followups include #144, #146, #147, #149, and eventually even #148.

## Parachain Core crate

The core of this PR is the `tuxedo-parachain-core` crate. It is an extention to the pre-existing `tuxedo-core` crate that adds types and methods relevant to the parachain protocol.

One design goal is that sovereign chain developers should not have to compile any of the heavy Polkadot or Cumulus crates. Therefore the parachain code is in a separate crate. The new crate has a re-export of the `tuxedo-core` crate. This allows parachain developers to depend only on `tuxedo-parachain-core` and not risk mis-matching versions.

### Validate Block function and `register_validate_block!` macro.

This PR also contains a working PoC of the `validate_block` function that will be used to validate parachain block on the relay chain validators. While the current implementation is good enough for testing, it still has a few security checks disabled, especially checking timestamp inherent, that need to be properly addressed before it is ready for any serious use.

There are some tricks about shared memory involved in the parahcains protocol that do not need to be exposed to parachain runtime developers. For this reason, we wrap the `validate_block` function's implementation that does the unsafe memory operations internally. This is the same in FRAME.

## Parachain Inherent Piece

Another addition is a Tuxedo piece in the wardrobe called the "parachain piece". This piece is modeled after the timestamp piece and it is responsible for handling the parachain inherent. This piece is atypical in one important way: it has side effects. Specifically, it writes the relay chain parent number to a fixed storage location. The effects are isolated by a clean mutator interface. Still this design is not 100% UTXO native, and better ideas are welcome in followups.

## Template Runtime

There is a new feature called `parachain` in the `tuxedo-template-runtime`. When this feature is enabled the runtime has the parachain piece and `validate_block` function included.

## Parachain Node

This PR is +11k lines of code, and a huge amount of that is in the parachain template node. This node is mostly copied directly from the upstream parachain template node in the SDK repo. This is the same approach we took in the sovereign node.

The only customized aspects are the inherent data providers, and the dev service.

## Dev Service

The dev service is a way to test the parachain runtime without a backing relay chain. This is a technique that I invented in Moonbeam in https://github.com/moonbeam-foundation/moonbeam/pull/204, https://github.com/moonbeam-foundation/moonbeam/pull/260, and a few others.

[...]

PR #139, opened 2023-11-13
renamed `SigCheck` to `Sr25519Signature`

### LINKED ISSUE #132: Standardize Verifier names
I noticed that `SigCheck` and `ThresholdMultiSignature` follow two fundamentally different naming standards.
I think we should either push for the long extensive name, converting the `SigCheck` (where I think the word "Check" is superfluous) to something like `SimpleSignature`, or change the second one to some ugly contraption like `ThreshMultiSig`, which is still well understandable but less elegant.

@JoshOrndorff said:
> I slightly prefer the long extensive names. [...] If you find it worth the effort afterwards, I trust your judgement to rename things, but I'm also fine leaving it as it is.

muraca:
closes #132


muraca:
I need to rebase on main after #130

JoshOrndorff (approved):
Well, I'm happy to see that this didn't require any special parachain effort. I was thinking we would have to change things for the sovereign chain and parachains.

PR #150, opened 2023-11-22
Dedicated parachain runtime

### LINKED ISSUE #116: POC Recursive tuxedo piece aggregation
> These slacks and jacket came together from my tailor. And this tie, belt, and shoes were a gift from my wife. Tonight I'm wearing the tailor's suit with my wife's accessories and this shirt I got from the thrift shop.

One goal I've had in mind for Tuxedo, which is probably not mission critical but would facilitate a lot of composability and also be really elegant, is the **ability to aggregate a Tuxedo `ConstraintChecker` out of constituent checkers that are themselves aggregated from their own constituents** :turtle: :turtle: :turtle: ...

When designing the aggregation macro I've tried my best to make sure that this recursive aggregation is possible, even going to great lengths during #100 to preserve the property. However I have **never actually tried it**.

This issue is to create a POC crate that contains a Tuxedo runtime using a recursively aggregated constraint checker, and the write some basic functional tests showing that it works as expected (even with inherents :crossed_fingers:).

The crux of the problem will be something like this pseudocode
```Rust
#[tuxedo_constraint_checker(Verifier)]
pub enum MoneyAndTime {
    /// Checks monetary transactions in a basic fungible cryptocurrency
    Money(money::MoneyConstraintChecker<0>),
    /// Set the block's timestamp via an inherent extrinsic.
    SetTimestamp(timestamp::SetTimestamp<Runtime>),
}

#[tuxedo_constraint_checker(Verifier)]
pub enum Biology {
    /// Checks Free Kitty transactions
    FreeKittyConstraintChecker(kitties::FreeKittyConstraintChecker),
    /// Checks that an amoeba can split into two new amoebas
    AmoebaMitosis(amoeba::AmoebaMitosis),
    /// Checks that a single amoeba is simply removed from the state
    AmoebaDeath(amoeba::AmoebaDeath),
    /// Checks that a single amoeba is simply created from the void... and it is good
    AmoebaCreation(amoeba::AmoebaCreation),
}

#[tuxedo_constraint_checker(Verifier)]
pub enum RecursivelyAggregatedRuntime {
    /// Business related transactions related to time and money
    MoneyAndTime(MoneyAndTime),
    /// Nature related transactions related to population regeneration
    Biology(Biology),
    
    // TODO Find a way to make business and nature coexist and interact holistically.
    // This comment is both a jab capitalism's handling of climate change and also a real programming task
    // For this to be really compelling, we need a way to expose configs for the intermediate checkers
    // and do things like use the business currency to pay for amoeba minting atomically.
    
    /// Upgrade the Wasm Runtime
    RuntimeUpgrade(runtime_upgrade::RuntimeUpgrade),
}
```

JoshOrndorff:
Closes #116 (if merged all the way to main)

This PR (which is currently based off of #130) introduces a dedicated parachain runtime instead of having a `parachain` feature in the existing runtime.

This solves a problem that when you simply ran `cargo build --release` in the repo, it would build the runtime with the parachain feature which meant the sovereign node would be broken.

But don't worry! This does not bring back (much) code duplication. We are able to take the sovereign runtime's constraint checker and use it as a constituent piece in the parachain's checker. This is all thanks to out aggregation macro's focus on recursion.

I included some comments to flex that our design is nicer than FRAME's in this regard.

One remaining issue is that the wallet can't decode the transactions properly (because the aggregate enum is one layer deeper and thus the encoding is one byte longer).


JoshOrndorff:
Closing in favor of #198

PR #152, opened 2023-11-23
added timestamp to wallet

muraca:
This is useful to check that #151 is working as expected.

### DIFF
--- a/wallet/src/sync.rs
+++ b/wallet/src/sync.rs
@@ -20,12 +20,13 @@ use sled::Db;
 use sp_core::H256;
 use sp_runtime::traits::{BlakeTwo256, Hash};
 use tuxedo_core::{
+    dynamic_typing::UtxoData,
     types::{Input, OutputRef},
     verifier::Sr25519Signature,
 };
 
 use jsonrpsee::http_client::HttpClient;
-use runtime::{money::Coin, Block, OuterVerifier, Transaction};
+use runtime::{money::Coin, timestamp::Timestamp, Block, OuterVerifier, Transaction};
 
 /// The identifier for the blocks tree in the db.
 const BLOCKS: &str = "blocks";
@@ -38,6 +39,9 @@ const UNSPENT: &str = "unspent";
 
 /// The identifier for the spent tree in the db.
 const SPENT: &str = "spent";
+
+/// The identifier for the current timestamp in the db.
+const TIMESTAMP: &str = "timestamp";
 
 /// Open a database at the given location intended for the given genesis block.
 ///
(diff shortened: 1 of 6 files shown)


wallet/src/sync.rs
```diff
@@ -262,23 +266,30 @@ async fn apply_transaction<F: Fn(&OuterVerifier) -> bool>(
...
-
-        match output.verifier {
-            OuterVerifier::Sr25519Signature(Sr25519Signature { owner_pubkey }) => {
-                // Add it to the global unspent_outputs table
-                add_unspent_output(db, &output_ref, &owner_pubkey, &amount)?;
+        // For now the wallet only supports simple coins and timestamp
+        match output.payload.type_id {
+            Coin::<0>::TYPE_ID => {
+                let amount = output.payload.extract::<Coin<0>>()?.0;
+
```
JoshOrndorff, line 273: This may be a good time to make the filtering we do here paramaterizeable like the verifier filter is above. Or at least make a follow up issue for it.

wallet/src/sync.rs
```diff
@@ -456,3 +467,12 @@ pub(crate) fn get_balances(db: &Db) -> anyhow::Result<impl Iterator<Item = (H256
...
 }
+
+pub(crate) fn get_timestamp(db: &Db) -> anyhow::Result<u64> {
+    let timestamp_tree = db.open_tree(TIMESTAMP)?;
+    let timestamp = timestamp_tree
+        .get([0])?
+        .ok_or_else(|| anyhow!("Could not find timestamp in database."))?;
+    u64::decode(&mut &timestamp[..])
+        .map_err(|_| anyhow!("Could not decode timestamp from database."))
+}
```
JoshOrndorff, line 478: Ideally this would go in a separate timestamp module. My idea is that logic associated with each piece (like how this is associated with the timestamp piece) will live in a dedicated file. That should make it easy to make chain-specific wallets by either including or not including that part of the wallet.

JoshOrndorff (approved):
Overall good. Made a couple little suggestions. Take them or leave them.

wallet/src/sync.rs
```diff
@@ -262,23 +262,15 @@ async fn apply_transaction<F: Fn(&OuterVerifier) -> bool>(
...
+        match output.payload.type_id {
+            Coin::<0>::TYPE_ID => {
+                crate::money::apply_transaction(db, tx_hash, index as u32, output)?;
+            }
+            Timestamp::TYPE_ID => {
+                crate::timestamp::apply_transaction(db, output)?;
             }
-            _ => return Err(anyhow!("{:?}", ())),
+            _ => continue,
         }
```
muraca, line 274: @JoshOrndorff  I like this approach better, please check it out.

JoshOrndorff (approved):
Thanks for improving the wallet structure!

Issue #155, opened 2023-11-28
Make CI run tests again


muraca:
Tests fail because after #130 disk space is not enough.

muraca:
Adding a `--release` flag was enough to make it work. https://github.com/Off-Narrative-Labs/Tuxedo/pull/158#issuecomment-1855473402

In #157 I wrote a custom GH action to split tests in different jobs, one for each section. Since Joshy prefers to have only one test action, I will close the PR but mention it here as well, so we can go back to it if we ever need it.

In #158 Joshy tried to use a GH action called `maximize-build-space`, but cargo gets stuck with some windows crates.

PR #157, opened 2023-11-30
improve CI tests

muraca:
This PR introduces a custom action to run tests for different crates, and updates the workflow to use this action for testing:
- `tuxedo-core`;
- `tuxedo-parachain-core` and `parachain-piece`, which is currently skipped as it fails because it uses too much disk space;
- `tuxedo-template-runtime`, which I think we could skip entirely;
- all the other wardrobe pieces, which should be updated manually when the list changes.

This PR is a step towards #155.

### DIFF
--- a/.github/actions/test/action.yml
+++ b/.github/actions/test/action.yml
@@ -6,10 +6,14 @@ branding:
   color: "green"
 
 inputs:
-  args:
-    description: "Arguments to pass to `cargo nextest run`."
+  packages:
+    description: "Packages to be tested by `cargo nextest`."
     required: true
     default: ""
+  ignore:
+    description: "Folders or files to be ignored by `cargo llvm-cov report`."
+    required: true
+    default: "target" # we need a default value, otherwise the command doesn't work
   title:
     description: "Title of the comment with the code coverage."
     required: true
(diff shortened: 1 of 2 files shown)


JoshOrndorff:
I'm open to going this route if it is necessary. But I have some hesitations.

1. It adds the maintenance overhead of a custom action. I'd rather stick closely to the simple recipes from https://github.com/actions-rs/example
2. It is kind of a workaround for the underlying problem which is that our code takes way too fing long to compile. Obviously we inherent this problem from Polkadot, but I'd like to try to improve the compile time first.

muraca:
> It adds the maintenance overhead of a custom action.

True to a certain extent, since it is like a code function we call from multiple sources. It has been made just to avoid copy-pasting the same thing over and over again.

>  I'd rather stick closely to the simple recipes from [actions-rs](https://github.com/actions-rs/example)

Well, `actions-rs` has been archived and is no longer maintained. But I don't see how it would be relevant for us.  
Do you mean we should find an "open source" action that does what we need to (eg test and code coverage using cache) and stick with that? I don't necessarily agree.

> I'd like to try to improve the compile time first.

Sorry if this wasn't clear, but this PR goes in an opposite direction: since we know where the problem with compile time and testing is, we make sure that everything else works and is tested as expected, otherwise we might overlook on some things and introduce bugs or other issues, or forget to test something because we don't check the coverage.

JoshOrndorff:
Okay, I'm coming around to this more and more.

> actions-rs has been archived and is no longer maintained.

Damn. I didn't realize that. That is a loss and I hope the community carries on that work somehow. I've found it very useful.

My point wasn't so much that we need to use that specific thing. My point was that I want the CI to be simple, straight-forward, idiomatic, etc. As opposed to custom, complex, hard-to-understand, etc.

We aren't CI experts (at least I'm not) and I don't really want to be one. I just want something simple that tells me if the tests pass. That's what I meant when I was talking about those actions-res examples.

I got the rocks-db dependency removed again over in #156 and I thought maybe that would save us enough disk space to not need this. But it wasn't enough.

muraca:
We might try to test parachain with the `--release` flag to see if it takes less space.

muraca:
superseded by #159

PR #158, opened 2023-12-13
Add maximize-build-space action

JoshOrndorff:
inspired by cdaa8c7bb385ca42359d0071348b2a7805dc55d3

This PR adds this simple magical (hacky) github action which gets us enough extra disk space to build the massive polkadot crates.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,8 @@ jobs:
       issues: write
       pull-requests: write
     steps:
+      # Polkadot crates are so bloaty, we need this hacky action to claim more disk space.
+      - uses: easimon/maximize-build-space@v10
       - uses: actions/checkout@v4
       - name: Install tooling
         run: |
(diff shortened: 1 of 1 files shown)


JoshOrndorff:
running `cargo clean && cargo nextest run --release && du -h` locally shows that my Tuxedo directory totals to 11G after all tests run. That seems like it should be good enough. Maybe even enough to fit in the 14GB that we have without the magic action.

Let's see what the disk usage is without the release flag.

muraca:
The action did not solve the problem we had, cargo gets stuck on some windows crate.
Thanks for testing my intuition with `--release` and pointing out it was indeed a good solution.

muraca:
The action did not solve the problem we had, cargo gets stuck on some windows crate.
Thanks for testing my intuition with `--release` and pointing out it was indeed a good solution.

JoshOrndorff:
> cargo gets stuck on some windows crate

How did you determine that. I can't see any logs from the failed jobs. (except for the first few steps which succeeded)

muraca:
> > cargo gets stuck on some windows crate
> 
> How did you determine that. I can't see any logs from the failed jobs. (except for the first few steps which succeeded)

I ran the action and looked at the logs in real time for 30 minutes. I don't recommend doing it.

PR #159, opened 2023-12-14
run tests with release flag

### LINKED ISSUE #155: Make CI run tests again

muraca:
run tests with release flag and beautify lines coverage print
closes #155


JoshOrndorff (approved):

JoshOrndorff:
Good step in the right direction. Now we know that the `--release` flag is enough, and that it completes in about an hour. But now we have some real error about `--exclude`. I guess this is from the command in the workflow.

PR #168, opened 2024-02-05
Hash Time Lock Contracts

JoshOrndorff:
This PR is a first step towards more interesting verifiers and cross-chain atomic swaps. Already there are enough verifiers to support swaps between two Tuxedo chains with the HTLC verifier :rocket:

Before we can realize the goal of serving on the arbitrating side of the Farcaser protocol we would need Either a verifier that supports bitcoin script, or dedicated farcaster verifiers, both of which are feasible and put off until a future PR.

Some references I followed:
* https://en.bitcoin.it/wiki/Hash_Time_Locked_Contracts
* https://lightning.network/lightning-network-paper.pdf Section 4
* https://arxiv.org/pdf/1801.09515.pdf Section 4.1
* https://www.youtube.com/watch?v=tcZGf8Qb_XA
* https://www.youtube.com/watch?v=8FeAXjkmDcQ

### New verifier implementations

* Simple hash lock. Can be spent simply by supplying a preimage.
* Simple time lock. Can be spent by anyone after a given block height.
* P2PKH. Private ownership just like a simple sig check but doesn't leak the pubkey. Inspired by bitcoin.
* HTLC. Has a spend path that requires a signature of the recipient and a preimage. Also has a refund path which requires a time delay and a signature by the refunder.


tuxedo-core/src/verifier.rs
```diff
@@ -8,44 +8,32 @@
...
+/// A means of checking that an output can be spent. This check is made on a
 /// per-output basis and neither knows nor cares anything about the validation logic that will
 /// be applied to the transaction as a whole. Nonetheless, in order to avoid malleability, we
 /// we take the entire stripped and serialized transaction as a parameter.
+///
+/// Information available when verifying an input includes:
+/// * The simplified transaction - a stripped encoded version of the transaction
+/// * Some environmental information such as the block current block number
+/// * An encoded redeemer supplied by the user attempting to spend the input.
+///   The redeemer is opaque to the trait and must be interpreted by the implementation.
```
JoshOrndorff, line 30:
```suggestion
/// * An redeemer supplied by the user attempting to spend the input.
```

JoshOrndorff: The redeemer now comes already-decoded :rocket:

tuxedo-core/src/verifier/htlc.rs
```diff
@@ -0,0 +1,380 @@
+//! This module contains `Verifier` implementations related to Hash Time Lock Contracts.
```
JoshOrndorff, line 1: For review purposes, this file is the main additions.

The rest of the diff is mostly moving existing verifiers to dedicated files.

tuxedo-core/src/verifier/htlc.rs
```diff
@@ -0,0 +1,380 @@
...
+    pub hash_lock: H256,
+    /// The pubkey that is intended to receive and acknowledge receipt of the funds.
+    pub recipient_pubkey: Public,
+    /// The time (as a block height) when the refund path opens up.
+    pub claim_period_end: u32,
+    /// The address who can spend the coins without revealing the preimage after the claim period has ended.
+    pub refunder_pubkey: Public,
+}
+
+///
```
muraca, line 97: I think you forgot some comments here :)

tuxedo-core/src/verifier/htlc.rs
```diff
@@ -0,0 +1,380 @@
...
+/// Allows a UTXO to be spent, and therefore acknowledged by an intended recipient by revealing
+/// a hash preimage. After an initial claim period elapses on chain, the UTXO can also be spent
+/// by the refunder. In practice, the refunder is often the same address initially funded the HTLC.
+///
+/// The receiver and refunder are specified as a simple public keys for simplicity. It would be
+/// interesting to use public key hash, or better yet, simply abstract this over some opaque
+/// inner verifier for maximum composability.
+///
+/// @lederstrumpf when the time expires, is the primary "happy" path supposed to remain open?
+/// Or is it that if the swap hasn't started on time, the refund is the only option.
```
muraca, line 84: I think these kinds of comments should be kept out of the docs

tuxedo-core/src/verifier/simple_signature.rs
muraca, line 1: I think we should keep the naming for this file coherent with the `Sr25519Signature` schema

muraca (changes requested):
I would have preferred to see the changes to the `Verifier` trait in a separate PR, then make HTLC depend on it. Although being small, I think it's a better approach for such a breaking change, but that's up to you.

Anyway the Redeemer is a really nice and clever addition, great job with that!
And the HTLC implementation looks good to me.

PS please take a look to the failing test 😿

tuxedo-core/src/verifier/htlc.rs
JoshOrndorff: Agreed. This one needs to be resolved before this is merged. Based Andreas's point in the video ([link to exact moment](https://youtu.be/8FeAXjkmDcQ?si=RkVmBBbaXo84WwAN&t=1492)) I believe that that the happy path should remain enabled which is how I have it now.

tuxedo-core/src/verifier/simple_signature.rs
JoshOrndorff: I was thinking more about the Sr25519Signature, and similar about the thing that might have been named `Sr25519BlakeTwo256HashLock`.

I think the right thing to do is keep the simple `Sr25519Signature` to show just how dead simple a verifier can be. But then make all the others generic over Substrate's notion of `Signature` and `Hash`. That way we don't Have to copy the logic when we someday need a `Ed25519Sha3HashLock` or similar.

tuxedo-core/src/verifier/htlc.rs
Lederstrumpf: > /// @lederstrumpf when the time expires, is the primary "happy" path supposed to remain open?
> /// Or is it that if the swap hasn't started on time, the refund is the only option.

Strong question you raise.
In the Farcaster protocol, the happy path closes once one of the parties initiates the cancel path.

HTLC atomic swap protocols on Bitcoin keep the happy path open even after timeout expires - see e.g. the COMIT spec: https://comit.network/docs/0.13.0/core-concepts/atomic-swap-htlc/#htlcs-on-different-ledgers (interestingly, they've done the same for the ETH protocol: https://github.com/comit-network/RFCs/blob/master/RFC-007-SWAP-Basic-Ether.adoc)
or the original proposal:
https://bitcointalk.org/index.php?topic=193281.msg2224949#msg2224949
or BIP 119:
https://github.com/bitcoin/bips/blob/master/bip-0199.mediawiki#summary

The reason for this is indeed that there's deliberately no `OP_CHECKSEQUENCE` nor `OP_CHECKLOCKTIME` - only their guardian variants `OP_CHECKSEQUENCEVERIFY` & `OP_CHECKLOCKTIMEVERIFY`.
Most discussion around this is probably in the annals of the bitcoin wizards IRC, if you wanna investigate deeper.
The arguments that have bubbled to my surface:
Satoshi's stance in 2010 (i.e. The Prophet's unassailable proclamation in Bitcoin Land, for better or worse): https://bitcointalk.org/index.php?topic=1786.msg22119#msg22119
Jeremy Rubin's AMA: https://www.reddit.com/r/Bitcoin/comments/gwjr8p/comment/ft80amd/?utm_source=share&utm_medium=web2x&context=3
Another disgruntled (& highly ignored) seeker of The Truth's summary: https://bitcoin.stackexchange.com/questions/96366/what-are-the-reasons-to-avoid-spend-paths-that-become-invalid-over-time-without

Now, while I have reservations about the reorg arguments even in Bitcoin's context, they are not applicable to chains with deterministic finality, i.e. all substrate based chains (aside from Kulupu).
i.e. with deterministic finality, you could have the equivalent of `OP_CHECK{SEQUENCE,LOCKTIME}`, and it would be safe as long as the output being spent has been finalized.

I reckon it might be preferable to have mutually exclusive success & refund paths for HTLC based swaps too - with a more general definition of HTLC, i.e. going beyond what Bitcoin Script allows.
It would still not ensure atomicity (since the chains run on different clocks), but improves safety of the swap for Bob assuming chain A's clock doesn't outrace chain B's clock.
Breakable assumption if crosschain: `t1 < t2`:
Let's take chain A to be the chain where Alice funds TX1 (with a timelock `t2` for her refund `TX2`) and chain B as the chain where Bob funds `TX3` (with a timelock `t1` for his refund `TX4`)

Assuming both funding txs have been published, but Alice has not claimed yet, then the possible spending conditions are:

`TX1` (Chain A): `(x ∧ Bob) ∨ (t2 ∧ Alice)`
`TX3` (Chain B): `(x ∧ Alice) ∨ (t1 ∧ Bob)`

In detail:
- `!t1 ∧ !t2`: Alice can claim. Bob can't do jack and has to wait for `t1` or Alice to claim.
- `t1 ∧ !t2` (`t1` has passed): Alice can claim (unsafe since Bob would learn `x` and his refund might frontrun her claim tx) or wait for `t2` to refund. Bob can (and should) refund or (unsafe!) wait for Alice to claim.
- `t1 ∧ t2` (both `t1` and `t2` have passed): Alice can claim or refund. Bob can refund and should already have done so before `t2` passed, else non-atomic.
- `!t1 ∧ t2` (`t2` has passed): Alice can claim or refund. Bob has to wait for `t1` or for Alice to claim: non-atomic scenario.

If you now impose that the happy path is unavailable after `t1` or `t2` respectively, then this changes to:
Chain A: `(!t2 ∧ x ∧ Bob) ∨ (t2 ∧ Alice)`
Chain B: `(!t1 ∧ x ∧ Alice) ∨ (t1 ∧ Bob)`

- `!t1 ∧ !t2`: Alice can claim. Bob can't do jack and has to wait for `t1` or for Alice to claim.
- `t1 ∧ !t2` (`t1` has passed): Alice can't claim (which would anyway be unsafe), so must wait for `t2` to refund. Bob can refund.
- `t1 ∧ t2` (both `t1` and `t2` have passed): Alice can only refund. Bob can only refund.
- `!t1 ∧ t2` (`t2` has passed): Alice can claim or refund. Bob has to wait for `t1` or for Alice to claim: non-atomic scenario.

The primary advantage is that Alice can't steal from Bob unless she publishes `TX2` prior to `t1`.
Note you'd get the same if you'd just make the paths on `TX1` temporally mutually exclusive, assuming Alice never leaks `x` to Bob (e.g. by hopelessly trying to spend `TX1`, or by getting hacked).
So if this was a crosschain swap of Tuxedo with e.g. Bitcoin, you could retain this advantage as long as `Tuxedo = Chain A`.

tuxedo-core/src/verifier/htlc.rs
```diff
@@ -0,0 +1,380 @@
...
+                // Check that the refunder has signed properly
+                sp_io::crypto::sr25519_verify(
+                    signature,
+                    simplified_tx,
+                    &self.refunder_pubkey,
+                )
+            }
+        }
+    }
+}
```
Lederstrumpf, line 138: What's your stance on composable `Verifier`?

I'm otherwise undecided whether you are deliberately not composing `HashLock` & `TimeLock` implementations to construct `HashTimeLockContract`, or are leaving this for a later stage.
I think there are pros (e.g. leaner interface & codebase) but also cons (more moving parts beneath a Verifier, so requires more careful versioning) to composing.
See https://github.com/Off-Narrative-Labs/Tuxedo/commit/29a20fe1 for a candidate implementation that reuses `Verifier` implementations from `HashLock` & `TimeLock`.

Lederstrumpf (commented):
lgtm so far :)

Left comments on the HTLC implementation where I reckon some decisions should be made.

tuxedo-core/src/verifier/htlc.rs
Lederstrumpf: TLDR: to keep compatibility with Bitcoin Script (not in terms of only supporting its features, but also the *assumptions* about what can happen), I'd keep the happy path open.

But keeping compatibility with BS's assumptions means restricting `Verifier` rules to what BS would permit, so this is a pretty fundamental decision. The other alternative is going with whatever makes most sense for a particular protocol and breaking compatibility with BS's assumptions. And that's also a pretty fundamental decision ;)

tuxedo-core/src/verifier/htlc.rs
JoshOrndorff: Thanks for the detailed writeup and citations regarding expiring transactions. I'm not all that convinced by the Prophet's argument TBH. As others explained in the forum, there are already double spend scenarios that make a transaction invalid from one fork to another, so I don't see expiry being that much worse. The state boat from expired transactions is more convincing to me, but in the case of swaps, the transaction never truly expires, only one of its paths does, so I'm not too concerned. I think for now we let the happy path close for the reasons you explained. Anyone could write their own verifier that leaves it open or adds a `happy_path_expires: bool` field if they want it later.

I think composing the verifiers is a good idea. I was already starting to consider it when I was writing this PR. Even looking at the simple `HashLock`, for example, even it could be composed with an inner `Sr25519Signature` rather than repeating the sig check logic. And by making them composable, we can also make them generic. I could imagine some verifier combinators for scenarios like `And` `Or` `Not` etc. I'll experiment with it.

tuxedo-core/src/verifier/htlc.rs
JoshOrndorff: Because Tuxedo is a framework, we don't have to make any decision once and for all here. The framework can support any old verifier whether it is compatible with bitcoin or not.

When it comes time to launch a Tuxedo-based chain then we have to decide what verifiers are concretely allowed. I don't think we should enforce bitcoin compatibility to the entire Tuxedo ecosystem.

tuxedo-core/src/verifier/htlc.rs
JoshOrndorff: Fixed in e9fb8d85d2e6

tuxedo-core/src/verifier/htlc.rs
```diff
@@ -0,0 +1,380 @@
...
+pub struct BlakeTwoHashLock {
+    pub hash_lock: H256,
+}
+
+impl BlakeTwoHashLock {
+    pub fn new_from_secret<T: Encode>(secret: T) -> Self {
+        Self {
+            hash_lock: BlakeTwo256::hash(&secret.encode()),
+        }
+    }
```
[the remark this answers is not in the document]
JoshOrndorff: Addressed in 551d5743bedd9e69749b808217ccb97ce095ecfe

JoshOrndorff:
This will be merged or rebased after #170 goes in. Then this will just be about the hashlock timelock and HTLC.

tuxedo-core/src/verifier/htlc.rs
muraca: I love the composed Verifier. Although it's not really more convenient in this particular case, it is actually a great example of what I believe should be a good practice in action.

PR #172, opened 2024-02-13
Improve Verifier Trait: Redeemer type and block number access

JoshOrndorff:
First, I've made a small change to the `verify` method where we also pass in the block number. This may seem like the beginning of a slippery slope about what data will be made available to verifiers. I think some good boundaries are block header data and stuff that is available to bitcoin scripts.

Second, we've added an associated type `Redeemer`. Previously we just accepted a slice of bytes to represent the redeemer. And the first thing that every redeemer did was decode it to some expected type. By having an associated type, the executive can do the decoding and handle errors. This makes life easier for `Verifier` writers. See the changes to `ThresholdMultisig` as an example of why the new interface is cleaner.


JoshOrndorff:
Thanks @muraca and @gorzorg for making the macro better. One other thing we could do is make the function name snake case with this crate https://crates.io/crates/convert_case. Perhaps here or in a follow up. I'm fine with either.

muraca (approved):

PR #173, opened 2024-02-13
Generic `SignatureCheck` `Verifier`

JoshOrndorff:
This PR introduces a new `Verifier` implementation called`SignatureCheck`. It is similar to the existing `Sr25519Signature` except that it is generic over Substrate's notion of crypto `Pair`.

### DIFF
--- a/tuxedo-core/src/verifier.rs
+++ b/tuxedo-core/src/verifier.rs
@@ -14,7 +14,7 @@ mod multi_signature;
 mod simple_signature;
 
 pub use multi_signature::ThresholdMultiSignature;
-pub use simple_signature::{Sr25519Signature, P2PKH};
+pub use simple_signature::{Sr25519SignatureCheck, P2PKH};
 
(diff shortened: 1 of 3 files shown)


muraca:
Am I missing something, or can we just ditch the `Sr25519SignatureCheck` struct and provide a `type Sr25519SignatureCheck = SignatureCheck< sr25519::Pair>;` with doc comments?

JoshOrndorff:
Yes, keeping the original `Sr25519SignatureCheck` is redundant at this point. I had originally left it there to show how you can easily use your own signature scheme even if it doesn't already exist in Substrate's code.

Maybe it makes more sense to do this for something like RSA https://lib.rs/crates/rsa which really doesn't exist in Substrate's code.

One potential problem is that the wallet uses sc-keystore, which only supports certain types of crypto not including RSA. So maybe the wallet just doesn't support rsa then; I guess that would be fine.

PR #188, opened 2024-03-01
Cleaner separation of `Verifier` / `Constraint Cheker` duties

JoshOrndorff:
The primary change this PR makes is to remove the `ConstraintChecker`'s access to the information about verifiers.

We added this feature thinking it was necessary for use cases such as sudo-style governance, royalties, and an on-chain treasury. It turns out, this was not the case and was more a reflection of our newness to thinking in the UTXO model. I've now demonstrated royalties in #185, sudo in #186, and sketched an [idea for a treasury](https://github.com/Off-Narrative-Labs/Tuxedo/pull/188#issuecomment-1973733858).

This change also simplifies the aggregation macro considerably which is always welcome, especially as I prepare to experiment with strong typing in the UTXO set a la #178.

A secondary change is that it cleans up the relationship between constraint checkers and inherents somewhat and removes the last hack from the existing inherent pieces. Inherents are still messier than I want but this is largely because Substrate's notion of inserting required info into block bodies via transactions is a little messy.


JoshOrndorff:
#### Treasury Design

Here is the sketch of a council-style treasury piece I mentioned in the pr description.

treasury-dao-lifecycle

muraca:
I was trying to figure out why it does not crash when `new_unspendable` is not implemented.

I tried to add a panic at the beginning of the Timestamp `create_inherent` function, but nothing happens.
Then I tried to add it at the beginning of this function https://github.com/Off-Narrative-Labs/Tuxedo/blob/6924966e382ab51c263e80527561f5cfd8e13e8a/tuxedo-core/src/inherents.rs#L176-L194 and nothing happens as well.

wardrobe/timestamp/src/lib.rs
```diff
@@ -249,7 +237,8 @@ impl<V: Verifier + From<UpForGrabs>, T: TimestampConfig + 'static> TuxedoInheren
 
         let new_output = Output {
             payload: new_timestamp.into(),
-            verifier: UpForGrabs.into(),
+            verifier: V::new_unspendable()
+                .expect("Must be able to create unspendable verifier to use timestamp inherent."),
```
JoshOrndorff, line 241: This is one panic I'm expecting to see when I run this node. But I don't see it.

wardrobe/timestamp/src/lib.rs
```diff
@@ -316,7 +305,9 @@ impl<V: Verifier + From<UpForGrabs>, T: TimestampConfig + 'static> TuxedoInheren
             peeks: Vec::new(),
             outputs: vec![Output {
                 payload: Timestamp::new(time, 0).into(),
-                verifier: UpForGrabs.into(),
+                verifier: V::new_unspendable().expect(
+                    "Must be able to create unspendable verifier to use timestamp inherent.",
+                ),
```
JoshOrndorff, line 310: Also this panic.

JoshOrndorff:
I solved the why-doesn't0it-panic mystery in [cc0758d](https://github.com/Off-Narrative-Labs/Tuxedo/pull/188/commits/cc0758d0a1d6aa20a4ddd789ab47c2df4ea115f9). I had forgotten to explicitly mark the inherents with the new `InherentAdapter` type.

JoshOrndorff:
One problem here is that the parachain node no longer works because the parachain inherent UTXO is now unspendable and can't be consumed by the next block's inherent. Some options:

1. Merge it as is with parachain node broken, and implement evictions next to fix it.
2. Change the parachain piece so it doesn't have to consume previous inherent, more like timestamp.
3. Switch the parachain piece back to abusing `UpForGrabs` for now.

Issue #98, opened 2023-09-27
Implement Evictions


JoshOrndorff:
Background: #52 

UPDATE:
I have pivoted from what I described in the original issue description, toward an earlier and simpler design as described in the [first comment](https://github.com/Off-Narrative-Labs/Tuxedo/issues/98#issuecomment-1926570968).

Original Description:

Currently the tuxedo executive checks the verifiers on all inputs of all transactions. We previously considered the notion of "eviction" which was a way for a transaction to consume a UTXO without its verifier being satisfied.

As I described in the [comment on the background issue](https://github.com/Off-Narrative-Labs/Tuxedo/pull/52#issuecomment-1736135134), I think a better approach is to make the notion of Verifier less fundamental in the tuxedo executive. Rather than always checking verifiers in the core, instead we never check verifiers there. Instead we generalize the notion of ConstraintChecker like we did in #41 so that there are three levels like follows.

```Rust
/// The simplest and most common way to express transactions logic.
/// Verifiers are checked for you automatically.
/// Only the input data is passed along to the implementer.
/// The same one we've had from the beginning
trait SimpleConstraintChecker { /* --snip-- */ }

/// The more expressive but still reasonably safe way to express transaction logic.
/// Verifiers are checked for you automatically.
/// The entire UTXO including the already-checked Verifier are passed along to the implementer.
/// The one that was introduced in #41
trait ConstraintChecker { /* --snip-- */ }

/// The most expressive and dangerous way to express transaction logic.
/// Nothing is checked for you.
/// Implementers get the raw input data and do as they please.
/// Checking Verifiers is entirely optional.
/// This one is new and should be used sparingly.
/// If you want to write an eviction, this is the way to do it.
trait FullPowerConstraintChecker { /*--snip--*/ }

// Two blanket implementations instead of the one we currently have.

```

JoshOrndorff:
Updates thoughts. Still similar to above. Add an enum

```rust
/// An input can be consumed in two way. It can be redeemed normally (probably with some signature, or proof) or it 
/// can be evicted. This enum is isomorphic to `Option<Vec<u8>>` but has more meaningful names
pub enum ConsumptionStrategy {
  /// The input is being consumed in the normal way with a signature or other proof provided by the spender.
  Redemption(Vec<u8>),
  /// The input is being forcefully evicted without satisfying its Verifier.
  Eviction,
```

Then change the input type:
```rust
pub struct Input {
[...]

PR #196, opened 2024-03-19
Implement evictions

JoshOrndorff:
Solves #98. Second try at evictions after #52.

"Normally" UTXOs are consumed in a process called "redemption" where the transacting user supplies some kind of signature or other proof that satisfies the UTXO's `Verifier`. An eviction is a (new alternative) way to forcefully consume input data without satisfying the verifier. In some usecases this is desirable. For example incentivizing cleaning up old state. Or to make governance decisions enforceable.

## Implementation

The changes revolve around this new enum:
```rust
/// An input can be consumed in two way. It can be redeemed normally (probably with some signature, or proof) or it 
/// can be evicted. This enum is isomorphic to `Option<Vec<u8>>` but has more meaningful names.
#[derive(Serialize, Deserialize, Encode, Decode, Debug, PartialEq, Eq, Clone, TypeInfo)]
pub enum RedemptionStrategy {
    /// The input is being consumed in the normal way with a signature or other proof provided by the spender.
    Redemption(Vec<u8>),
    /// The input is being forcefully evicted without satisfying its Verifier.
    Eviction,
}
```

This does not mean that any transaction can evict arbitrary storage. The ConstraintChecker will be informed about which inputs were evicted and it will enforce that evictions are only allowed in specific contexts (such as removing old votes after an election has been tallied).

## Relationship with Peeks

Currently (both before and after this PR) all peeks are evictions. Peeks are evictions in the sense that anyone can peek at any data without needing to satisfy its verifier.

Now that I am focusing on evictions, it is clear that there are cases where it would be nice to require a redemption to peek at a UTXO. For example, this would make #186 much more parallelizable. I'll make a followup issue for this, but I don't see it as priority for now.

## Must my Chain Allow Evictions?

No. In some chains, especially those serving communities strongly value property rights, evictions may never be desirable, and that is totally fine. It is straightforward to build a Tuxedo chain that has no eviction paths. But as a _framework_ Tuxedo also supports chains that want to support evictions


JoshOrndorff:
Status update. This compiles, but the parachain node still doesn't run. Haven't tried the standalone node yet. I'll do a clean compile and try again.

Issue #146, opened 2023-11-22
Proper general way to extract parachain inherent data


JoshOrndorff:
Our validate block implementation relies on a function called `extract_parachain_inherent_data`.

Currently this function assumes that the parachain inherent will be first. This should be improved so that it can be anywhere in the block.

The trick is that we need some way to determine whether we have the right constraint checker variant even though the aggregate constraint checker is generic. Perhaps something like [`IsSubType`](https://paritytech.github.io/polkadot-sdk/master/frame/traits/trait.IsSubType.html) would be useful.

JoshOrndorff:
Solved in #199

PR #199, opened 2024-04-02
`parachainify!` macro

JoshOrndorff:
This PR cleans up some hacks that were left when I first added parachain support. In particular it strives to solve #146.

There were several incidental changes as well, all of which I consider good:
* Persist block number in storage under new dedicated key.
* Remove generic block parameter from Executive. It is now just `Executive<V, C>`.
* Create and use some type aliases like `Header` and `Block<V, C>` in core.
* Reduce definitions in parachain runtime, instead re-using more stuff from the original template.


JoshOrndorff:
Node template and Parachain template cannot handle the example spend coin command in `--dev` mode. Gotta see when that got introduced.

```
$ tuxedo-template-wallet --dev --parachain \
  spend-coins \
  --output-amount 40 \
  --output-amount 50
```

